PR #8, opened 2015-06-19
Enabling PWM on pins A1/A2

cmaglie:
This PR enables PWM on pins A1/A2, but this sacrifice timer TC4 that is also used on Servo library, so using PWM on those pins conflict with Servo.

There are also some missing definitions on WVariant costants that has been added.


cmaglie:
rebased and merged

PR #4, opened 2015-06-17
add rtos systick entrance

Makeblock-official:
add rtos systick entrance for freertos
the free rtos porting placed at https://github.com/Makeblock-official/FreeRTOS-Arduino

PR #13, opened 2015-06-26
Fixed cortex-M hooks for RTOS

cmaglie:
Added RTOS handler like in the SAM core.

Fixes #4


cmaglie:
See also: http://forum.arduino.cc/index.php?topic=332080.0

PR #21, opened 2015-07-19
Added function setClock on bus I2C

PaoloP74:
Tested 100kHz and 400kHz speed on external EEPROM 24LC256 with extEEPROM library (https://github.com/PaoloP74/extEEPROM)
It should be tested with other devices.


cmaglie:
@PaoloP74 why you set the clock only for 100k and 400k?

is `sercom->initMasterWIRE(baudrate);` working only for these speeds?

PaoloP74:
No.
On datasheet (http://www.atmel.com/Images/Atmel-42181-SAM-D21_Datasheet.pdf) page 511+, master and slave I2C bus support 4 speed:
Standard-mode: 100khz
Fast-mode: 400khz
Fast-mode Plus: up to 1Mhz
High-speed mode: up to 3.4Mhz.
But the most used are 100kHz and 400kHz. I do not know device that go faster.

PR #23, opened 2015-07-25
Fix for deadlock conditions after i2c bus errors

pixelpixi:
Hello. I'm working on getting my project, Modulo (www.modulo.co) working with the Arduino Zero. Modulo uses hot plugging of i2c devices, so it tends to stress test handling of i2c bus errors. I found two places in SERCOM.cpp where incorrect handling of error conditions can cause a deadlock to occur. This pull request includes simple fixes for both of them.

The first occurs when an addressed device simply doesn't respond. This can be easily reproduced by doing a Wire.requestFrom() to an address that doesn't exist. Without the fix the arduino will completely lock up. With the fix, requestFrom returns an error. The bug is that it waits until the SB (slave on bus) bit is set in the INTFLAGS register, but when a nack occurs that never happens so we're stuck in an infinite loop. The fix is to also look for the MB flag to be set. If it is, issue a stop condition and return.

The second happens when a bus error (ie, an illegal stop condition) occurs while sending data as a master. This can be reproduced by connecting/disconnecting the SDA and SCL lines while i2c communication is in progress. In that case we are waiting for the MB (master on bus) flag to be set. When a bus error occurs that never happens, so again we end up in an infinite loop. The fix here is to also look for the BUSERR flag to be set. If it is, return an error condition.

Please let me know if there's any additional information I can provide on the problems or the fixes. Thanks.

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -493,6 +493,12 @@ bool SERCOM::startTransmissionWIRE(uint8_t address, SercomWireReadWriteFlag flag
   {
     while( !sercom->I2CM.INTFLAG.bit.SB )
     {
+        // If the slave NACKS the address, the MB bit will be set.
+        // In that case, send a stop condition and return false.
+        if (sercom->I2CM.INTFLAG.bit.MB) {
+            sercom->I2CM.CTRLB.bit.CMD = 3; // Stop condition
+            return false;
+        }
       // Wait transmission complete
     }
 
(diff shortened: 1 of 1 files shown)


facchinm:
Hi @pixelpixi, thank you for the patch! I've tested on a Zero and it works great.
I believe it will be merged as soon as @cmaglie is back.

pixelpixi:
Thank you for testing, @facchinm!

pixelpixi:
Hello @cmaglie and @facchinm. I was just wondering if you have an estimate for when this might get pulled in? If there are any concerns about it or anything else you need, please let me know what I can do. Of course if you are just busy and haven't gotten to it I completely understand.

Thanks a lot!

cmaglie:
Thank you!

pixelpixi:
Thrilled to see this get pulled in! Thanks!

PR #26, opened 2015-08-14
Relax digitalWrite parameter check

facchinm:
calling digitalWrite with any value different from 0 will end in performing digitalWrite(HIGH)
(to make it compatible with existing implementations for other cores)

### DIFF
--- a/cores/arduino/wiring_digital.c
+++ b/cores/arduino/wiring_digital.c
@@ -86,11 +86,8 @@ void digitalWrite( uint32_t ulPin, uint32_t ulVal )
       PORT->Group[g_APinDescription[ulPin].ulPort].OUTCLR.reg = (1ul << g_APinDescription[ulPin].ulPin) ;
     break ;
 
-    case HIGH:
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Looks good to me. The only reason this would be bad if we ever want to support a third parameter to digitalWrite (like `digitalWrite(HALF)` or something unexpected like that), but I think that would break too many sketches out here to work), so no practical downsides to this change.

aethaniel:
@facchinm, just for fun, this isn't matching the spec from https://www.arduino.cc/en/Reference/DigitalWrite

matthijskooijman:
But it does match other implementations and, I think, user expectation. Perhaps this expectation should be solidified by updating the reference, though.

aethaniel:
Indeed, the reference must be updated as it can be considered as the base spec for any new implementation. Well, it has been for me for the Due and Zero ports.

cmaglie:
@aethaniel 
to be precise, the new implementation match the specification exactly as the old one: the pin goes high when you pass `HIGH` and low when you pass `LOW`.

The specification leaves **undefined** all the other cases but the majority of implementations sets the pin high (I guess it's a consequence of the fact that in C/C++ `0` is considered `false` and everything else is `true`).

PR #27, opened 2015-08-19
USB Device improvements and stability fixes

cmaglie:
This pull request contains a clean up of the USB Device stack and fixes for a bunch of bugs:

- Source code has been rationalized by removing duplicate code scattered among `samd21_device.c`, `samd21_device.h` and `USB_device.h`.
- Fixed a bug in blocking-send buffering for USBDevice
- Fixed a bug in HID device descriptor handling
- Stability improvement in back-to-back SerialUSB communication

/cc @facchinm 
/cc @aethaniel

PR #35, opened 2015-09-07
PluggableUSB complete port

facchinm:
should apply on the top of PR #27


cores/arduino/cortex_handlers.c
```diff
@@ -176,3 +175,16 @@ void SysTick_Handler(void)
     return;
   SysTick_DefaultHandler();
 }
+
+static void (*usb_isr)(void);
+
+void USB_Handler(void)
+{
+  if (usb_isr)
```
aethaniel: why not putting the vectors in RAM instead of using this trick?
SCB->VTOR=new vectors address

PR #42, opened 2015-09-23
Fix warning on file stream.cpp

PaoloP74:
This fix eliminate a warning due to wrong data type used.
> Stream.cpp:167:10: warning: comparison is always false due to limited range of data type [-Wtype-limits] if(c < 0)


PaoloP74:
Fix with commit https://github.com/arduino/ArduinoCore-samd/commit/eb397e2b4d6916c53ab47ed6627a0eb3ed2f4177

PR #45, opened 2015-09-28
A fix for the SerialUSB buffer overflow bug.

GabrielNotman:
Send the data in chunks based on the buffer size
instead of assuming that the whole send data
will fit into the buffer.

This is one possible fix for the issue raised here:
https://forum.arduino.cc/index.php?topic=340897.0

### DIFF
--- a/cores/arduino/USB/USBCore.cpp
+++ b/cores/arduino/USB/USBCore.cpp
@@ -120,12 +120,32 @@ uint32_t USBD_Send(uint32_t ep, const void* d, uint32_t len)
 	int r = len;
 	const uint8_t* data = (const uint8_t*)d;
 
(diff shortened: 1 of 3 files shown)


GabrielNotman:
I believe this issue has now been resolved in commit:
ba6628574739608fbd94748db7469c5ff7c537a4

cmaglie:
Hi @GabrielNotman,

first of all, sorry for the lack of feedback, This is a very busy period and we failed to leave a comment...

> I believe this issue has now been resolved in commit

Yes, we were working on a bunch of USB fix, and your patch was basically already included in them, BTW it was a good fix for a compelling problem, so thank you anyway for your contribution!

PR #46, opened 2015-09-29
Wire Slave

sandeepmistry:
Some changes to get Wire slave working, including:
 * calling ``pinPeripheral`` in ``begin`` to setup SDA and SCL pins like the master ``begin``
 * enabling the stop condition interrupt (``SERCOM_I2CS_INTENSET_PREC``)
 * ACK'ing or NACK'ing in the ISR and sending the appropriate command bits to clear the flags

__Note:__ there is still an issue with repeated starts, the timeout is not working, so if a repeated start is received without a subsequent stop or another repeated start, the receive callback is not called currently.


cmaglie:
@ArduinoBot build this please

PR #47, opened 2015-10-10
Bootloader rework

aethaniel:
This patch provides a code cleanup and improvement of USB stability through better clock settings


cmaglie:
Thanks @aethaniel !

PR #65, opened 2015-11-17
Port some AVR Serial_ (SerialUSB) API's over

sandeepmistry:
Port over the following ```Serial_``` API's over from the AVR core to SAMD:

* ``readBreak``
* ``baud``
* ``stopbits``
* ``paritytype``
* ``numbits``
* ``dtr``
* ``rts``

That were added to AVR in the following commits:
* https://github.com/arduino/Arduino/commit/96f7b966bccd8b09acdd53a977183368983fb8e1
* https://github.com/arduino/Arduino/commit/b58f2394005cd073d582b8f90f8ba222aa95b0ee

cc/ @matthijskooijman

### DIFF
--- a/cores/arduino/USB/CDC.cpp
+++ b/cores/arduino/USB/CDC.cpp
@@ -58,6 +58,8 @@ static volatile LineInfo _usbLineInfo = {
 	0x00    // lineState
 };
 
+static volatile int32_t breakValue = -1;
+
 static CDCDescriptor _cdcInterface = {
 	D_IAD(0, 2, CDC_COMMUNICATION_INTERFACE_CLASS, CDC_ABSTRACT_CONTROL_MODEL, 0),
 
@@ -132,6 +134,12 @@ bool CDC_Setup(USBSetup& setup)
 			{
 				cancelReset();
 			}
+			return false;
+		}
+
+		if (CDC_SEND_BREAK == r)
+		{
+			breakValue = ((uint16_t)setup.wValueH << 8) | setup.wValueL;
 			return false;
 		}
 	}
@@ -287,6 +295,50 @@ Serial_::operator bool()
 
 	delay(10);
 	return result;
+}
+
+int32_t Serial_::readBreak() {
+	uint8_t enableInterrupts = ((__get_PRIMASK() & 0x1) == 0);
+
+	// disable interrupts,
+	// to avoid clearing a breakValue that might occur 
+	// while processing the current break value
+	__disable_irq();
+
+	int32_t ret = breakValue;
+
+	breakValue = -1;
+
+	if (enableInterrupts) {
+		// re-enable the interrupts
+		__enable_irq();
+	}
+
+	return ret;
+}
+
+unsigned long Serial_::baud() {
+	return _usbLineInfo.dwDTERate;
+}
+
+uint8_t Serial_::stopbits() {
+	return _usbLineInfo.bCharFormat;
+}
+
+uint8_t Serial_::paritytype() {
+	return _usbLineInfo.bParityType;
+}
+
+uint8_t Serial_::numbits() {
+	return _usbLineInfo.bDataBits;
+}
+
+bool Serial_::dtr() {
+	return _usbLineInfo.lineState & 0x1;
+}
+
+bool Serial_::rts() {
+	return _usbLineInfo.lineState & 0x2;
 }
 
 Serial_ SerialUSB(USBDevice);
--- a/cores/arduino/USB/USBAPI.h
+++ b/cores/arduino/USB/USBAPI.h
@@ -126,6 +126,46 @@ public:
 	virtual size_t write(const uint8_t *buffer, size_t size);
 	using Print::write; // pull in write(str) from Print
 	operator bool();
+
+	// This method allows processing "SEND_BREAK" requests sent by
+	// the USB host. Those requests indicate that the host wants to
+	// send a BREAK signal and are accompanied by a single uint16_t
+	// value, specifying the duration of the break. The value 0
+	// means to end any current break, while the value 0xffff means
+	// to start an indefinite break.
+	// readBreak() will return the value of the most recent break
+	// request, but will return it at most once, returning -1 when
+	// readBreak() is called again (until another break request is
+	// received, which is again returned once).
+	// This also mean that if two break requests are received
+	// without readBreak() being called in between, the value of the
+	// first request is lost.
+	// Note that the value returned is a long, so it can return
+	// 0-0xffff as well as -1.
+	int32_t readBreak();
+
+	// These return the settings specified by the USB host for the
+	// serial port. These aren't really used, but are offered here
+	// in case a sketch wants to act on these settings.
+	uint32_t baud();
+	uint8_t stopbits();
+	uint8_t paritytype();
+	uint8_t numbits();
+	bool dtr();
+	bool rts();
+	enum {
+		ONE_STOP_BIT = 0,
+		ONE_AND_HALF_STOP_BIT = 1,
+		TWO_STOP_BITS = 2,
+	};
+	enum {
+		NO_PARITY = 0,
+		ODD_PARITY = 1,
+		EVEN_PARITY = 2,
+		MARK_PARITY = 3,
+		SPACE_PARITY = 4,
+	};
+
 private:
 	USBDeviceClass &usb;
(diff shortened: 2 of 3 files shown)


matthijskooijman:
Just had a quick look over this commit, which looks good. Just one atomicity issue to fix.

sandeepmistry:
@matthijskooijman I've address the atomicity issue by disabling and re-enabling USB interrupts in ```readBreak``` (https://github.com/sandeepmistry/ArduinoCore-samd/commit/de1429bbbb7d074cc69d8ef88e9b270f6cd453c2).

Again as mentioned in https://github.com/arduino/Arduino/pull/4171#issuecomment-169796982, any ideas on how to test this condition?

matthijskooijman:
@sandeepmistry, two more remarks:
 - This always enables the interrupt, even if it was disabled before. I don't suppose it will ever be  disabled with the current code, but you never know what a sketch will do. In any case, it's good practice to clear the interrupt flag, and then restore it afterwards, instead of forcing it to enabled.
 - Are you sure you're disabling the right interrupt? Tracing back, I see the ISR routine [is called `UDD_Handler`](https://github.com/sandeepmistry/ArduinoCore-samd/blob/de1429bbbb7d074cc69d8ef88e9b270f6cd453c2/cores/arduino/USB/USBCore.cpp#L37), not [`USB_Handler`](https:/https://github.com/sandeepmistry/ArduinoCore-samd/blob/de1429bbbb7d074cc69d8ef88e9b270f6cd453c2/cores/arduino/USB/USBCore.cpp#L897). I'm just guessing here, I don't really know about the SAM register names.

As for testing this, one trick is to put a delay between the read and write of `breakValue`, which increases the chances of the race condition triggering. Furthermore, when a few  break requests are sent close together, you can miss some of them by not checking frequently enough, but you should *always* see the last one sent. If the race condition triggers, you can actually miss the most recent one.

To test, you could send a "break on" immediately followed by a "break off" and repeat that over and over again (with a bit of delay in between, so you can be sure that `readBreak()` was called). If you then let the break condition pull a line low, you should never see it low during the delay. If you do see it low for more than a short pulse, then you'll know you've missed a "break off" and the race condition triggered.

sandeepmistry:
@matthijskooijman thanks for reviewing and the tips on how to test!

I've addressed your feedback in https://github.com/sandeepmistry/ArduinoCore-samd/commit/d84826461fe1b30907fe61d00af2e620f130471d, by taking a similar approach to ``` Serial_::accept``` and using ```noInterrupts``` and ```interrupts``` to make the block atomic.

matthijskooijman:
@sandeepmistry, I actually think this change makes things worse (also, I think the implementation of `accept()` is also bad, my previous comment also applies there). Now, you disable *all* interrupts, instead of just the one that is needed (which is commonly done, but in this case not needed). More importantly, `interrupts()` still unconditionally enables interrupts, causing a call to `readBreak()` (or `accept()`) to (globally) *enable* interrupts if they were disabled. In that sense, things are worse now, since the chance that these functions are called with the global interrupt flag cleared are a lot bigger than with just the USB interrupt flag cleared. In any case, these should just store the state of the interrupt flag and restore it afterwards (ideally just for the USB interrupt, though disabling interrupts globally also prevents issues when e.g. `readBreak()` is called from *within* an ISR). I tried to look for an example of this clear-and-restore behaviour in some other code, but couldn't find anything quickly.

sandeepmistry:
@matthijskooijman good points, I think ```SPI:usingInterrupt``` has the approach we need: https://github.com/sandeepmistry/ArduinoCore-samd/blob/master/libraries/SPI/SPI.cpp#L100

It only calls ```interrupts()``` if ```__get_PRIMASK()```/```interruptsStatus``` was true. What do you think?

matthijskooijman:
Yup, that is indeed to solution I was aiming for. I'm not sure if it's worthwile to only disable the USB interrupt, or disable interrupts globally, though. The latter is safer, but only in corner cases that will probably never occur, and causes extra interrupt delay in other cases. But since the delay is minimal, I think I'd favor the global disabling.

On another note, never knew about the `interruptsStatus()` function, seems its samd (or also sam?) specific and undocumented?

matthijskooijman:
Oh, just noticed that that function is defined by SPI.cpp itself: https://github.com/sandeepmistry/ArduinoCore-samd/blob/master/libraries/SPI/SPI.cpp#L85-L93

sandeepmistry:
I'll move ```interruptsStatus()``` into a shared place in the SAMD core.

matthijskooijman:
Note that `interruptsStatus()` is probably fine to use in internal SAMD code, but I do not think it works well enough as a generally available API due to portability issues. There has been a thread about pretty much this feature a while ago, without any real conclusion. See https://groups.google.com/a/arduino.cc/d/msg/developers/cmu0Qy32zxY/PIo4YWtmAwAJ

sandeepmistry:
Another changed pushed, I've decided to use ```__get_PRIMASK``` directly instead of moving ```interruptsStatus``` from the SPI library into the SAMD core.

matthijskooijman:
That sounds good to me. I dug in a little more (to verify that "PRIMASK" is really the same bit as the one modified by `interrupts()/noInterrupts()`, and I found that the latter calls `__disable_irq()` which is defined in `~/.arduino15/packages/arduino/tools/CMSIS/4.0.0-atmel/CMSIS_RTX/SRC/rt_HAL_CM.h` as:
```
__attribute__((always_inline)) static inline U32 __disable_irq(void)
{
  U32 result;

  __asm volatile ("mrs %0, primask" : "=r" (result));
  __asm volatile ("cpsid i");
  return(result & 1);
}
```

In other words, `noInterrupts()` already returns the interrupt mask, so it might be even easier to use that (it's not portable, but this code does not need to be). To emphasize the non-portability (and reduce confusion for people reading the code), it might be better to even just use `__disable_irq()` and `__enable_irq()` directly?

Also note that the above code limits to just looking at bit 0 and masks out the rest, while the current commit (and SPI code) look at all bits. According to [the documentation] all bits but 0 are reserved, so this is probably not a problem in practice, but masking would be appropriate.

In other words:

sandeepmistry:
It turns out ```__disable_irq ``` from ```~/.arduino15/packages/arduino/tools/CMSIS/4.0.0-atmel/CMSIS_RTX/SRC/rt_HAL_CM.h``` is not included.

Instead the one from ```~/Library/Arduino15/packages/arduino/tools/CMSIS/4.0.0-atmel/CMSIS/Include/core_cmFunc.h``` is included. Which has a void return type.

```c
/** \brief  Disable IRQ Interrupts

  This function disables IRQ interrupts by setting the I-bit in the CPSR.
  Can only be executed in Privileged modes.
 */
__attribute__( ( always_inline ) ) __STATIC_INLINE void __disable_irq(void)
{
  __ASM volatile ("cpsid i" : : : "memory");
}
```

> Also note that the above code limits to just looking at bit 0 and masks out the rest, while the current commit (and SPI code) look at all bits. According to the documentation all bits but 0 are reserved, so this is probably not a problem in practice, but masking would be appropriate.

I've updated to mask.

sandeepmistry:
Here's a link to the relevant M0 documentation for my comment above: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0497a/BABEDCBC.html

> To emphasize the non-portability (and reduce confusion for people reading the code), it might be better to even just use __disable_irq() and __enable_irq() directly?

Agreed, I've pushed changes for this.

matthijskooijman:
> Instead the one from ~/Library/Arduino15/packages/arduino/tools/CMSIS/4.0.0-atmel/CMSIS/Include/core_cmFunc.h is included. Which has a void return type.

Ah, good point. I just looked at the first one I found using grep :-)

The code looks good now. Before merging, it would probably be good to squash all commits into one? Also, it might be good to let `accept()` use the same conditional `interrupts()` call (in a separate commit, of course)? Bit off-topic for this PR, though it would be good to change it anyway.

sandeepmistry:
> The code looks good now. Before merging, it would probably be good to squash all commits into one? 

Done, I've squash and rebased.

> Also, it might be good to let accept() use the same conditional interrupts() call (in a separate commit, of course)? Bit off-topic for this PR, though it would be good to change it anyway.

I'll open a new PR for this today.

matthijskooijman:
Looks good to me!

PR #64, opened 2015-11-13
[bootloader] [test needed] Bootloader leds activity

aethaniel:
This PR brings:
- activity LEDs support
- Genuino Zero build
- Arduino IDE 1.6.6 (MSWin specific change in APPDATA from Roaming to Local)

Readme.md updated with added 'how-to build' section.

I currently have a problem with my laptop which is refusing to mount properly the Arduino Zero.
Help is needed to verify the LED activity behaviour :-)

PR #69, opened 2015-12-04
Add Leonardo-like led pulse when bootloader is active

facchinm:
Use a naive way to produce a led "pulse", no pwm, only systick and unsigned arithmetics :smile: 
Based on top of #64


cmaglie:
Superseded by #165

Issue #28, opened 2015-08-23
the pin mode is changed to INPUT mode


Makeblock-official:
on file wiring.c line 87(Fun init()) Setup all pins (digital and analog) in INPUT mode (default is nothing) 
Pin mode should not be set.This line should comment.

on file wiring_digital.c line 81 (Fun digitalWrite()) Enable pull-up resistor.
This line should comment.

The pin mode will be initialized when construct the object.

For example:

```C++
class LED
{
  public:
    LED(int pin);
    void low();
    void high();
  private:
    int led_pin;
};

LED::LED(int pin)
{
  led_pin = pin;
  pinMode(led_pin, OUTPUT) ;
}

void LED::low()
{
  digitalWrite(led_pin, LOW);
}
void LED::high()
{
  digitalWrite(led_pin, HIGH);
}

LED led(13);
void setup() {
}

void loop() {
  led.high();
  delay(1000);
  led.low();
  delay(1000);
}
```

Since the pin mode is changed to INPUT,led.high() is set to input pull-up and led.low() is set to input pull-down.So the led will be very dark.

aethaniel:
Indeed, the pull-up should be set only on a digitalWrite(pin, HIGH) when pin is in INPUT mode.

Makeblock-official:
It is not correct, I mean, Pin mode should not be set at initialization.

the Fun void init( void )
  // Setup all pins (digital and analog) in INPUT mode (default is nothing)
  for ( ul = 0 ; ul < NUM_DIGITAL_PINS ; ul++ )
  {
	  pinMode( ul, INPUT ) ;//This line should comment.
  }

void digitalWrite( uint32_t ulPin, uint32_t ulVal )
{
  // Handle the case the pin isn't usable as PIO
  if ( g_APinDescription[ulPin].ulPinType == PIO_NOT_A_PIN )
  {
    return ;
  }

the Fun void digitalWrite( uint32_t ulPin, uint32_t ulVal )
  // Enable pull-up resistor
  PORT->Group[g_APinDescription[ulPin].ulPort].PINCFG[g_APinDescription[ulPin].ulPin].reg=(uint8_t)(PORT_PINCFG_PULLEN) ;//This line should comment.

Makeblock-official:
It is not correct, I mean, Pin mode should not be set at initialization.

the Fun void init( void )
  // Setup all pins (digital and analog) in INPUT mode (default is nothing)
[...]

PR #71, opened 2015-12-11
 fix issue #28  and some examples do not work    because main.cpp cal…

xiaoyu-makeblock:
…l void init( void ). Setup all pins (digital and analog) in INPUT mode (default is nothing)

such as :LiquidCrystal examples


sandeepmistry:
I'm closing this, as setting all pins to INPUT is desired, to maintain compatibility AVR based boards.

As mentioned by @WestfW in https://github.com/arduino/ArduinoCore-samd/issues/28#issuecomment-167461648.

PR #75, opened 2015-12-17
Fixed description send function to have working USB device on Windows

agdl:
With the previous implementation only the first byte of the USB description
was sent and so the driver result broken


sandeepmistry:
@agdl I'm closing this in favour of #102 for now ...

PR #77, opened 2015-12-18
Fixing potential xxxtoa() issue with newer ARM GCC toolchain

aethaniel:
This patch fix potential issue coming with newlib integrated at least since ARM GCC 4.9 2015q2:

https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=commit;h=32c96ddd14acadc82e7fccf110367ec3c8320c85

Some tests have been done without breaking anything at the moment.
Only WString.cpp is using this API.

Signed-off-by: Thibaut VIARD <aethaniel@sam-geek.org>

### DIFF
--- a/cores/arduino/itoa.c
+++ b/cores/arduino/itoa.c
@@ -17,7 +17,7 @@
 */
 
 #include "itoa.h"
-#include <string.h>
+#include <stddef.h> // For NULL
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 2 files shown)


sandeepmistry:
@aethaniel thanks for submitting!

Could you please expand on the potential issue?

I think we have bigger problems, as [Arduino.h](https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/Arduino.h) includes both ```<stdlib.h>``` and ```"itoa.h"``` so there will still be a conflict.

Would a better solution be to wrap some of the contents of ```itoa.h``` and ```itoa.c``` around macros based on the newlib version? So, the new built-in ```itoa``` and ```utoa``` functions can be used?

sandeepmistry:
I'm going to close this for now, we can revisit when the GCC version for the SAMD core is updated.

PR #78, opened 2015-12-21
add support for I2C repeated start, tested with MB85RC256V

ladyada:
repeated start on the atsamd21 is a little interesting, instead of telling the i2c master to send a REPEATED_START (which you would *think* is the thing to do) you should send **no** command. then, next time the ADDR is set, it will automatically do the repeated start for you.

See table 27-12 in datasheet:
***If CMD 0x1 is issued, a repeated start will be issued followed by the transmission of the current address in ADDR.ADDR. If another address is desired, ADDR.ADDR must be written instead of the CMD bits. This will trigger a repeated start followed by transmission of the new address.***

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -485,17 +485,19 @@ void SERCOM::prepareCommandBitsWire(uint8_t cmd)
   }
 }
 
-bool SERCOM::startTransmissionWIRE(uint8_t address, SercomWireReadWriteFlag flag)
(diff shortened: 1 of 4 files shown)


ladyada:
oops, already in https://github.com/arduino/ArduinoCore-samd/pull/66 - well now you have 2 versions :D

sandeepmistry:
@ladyada no worries, would you mind trying out #66?

ladyada:
tried git@github.com:sandeepmistry/ArduinoCore-samd.git -b "wire-master-repeated-start" and it also works just fine with the MB85RC256V

sandeepmistry:
@ladyada great, thanks for trying it out!

cmaglie:
Fixed with #66

PR #87, opened 2016-01-06
USB keyboard right shift key was ignored

bbx10:
The right shift key on a USB keyboard was ignored. The patch ensures the left and right shift keys are handled equally.


Lauszus:
Btw the same fix was merged into 2.0 some time ago: https://github.com/felis/USB_Host_Shield_2.0/pull/98.

PR #88, opened 2016-01-08
Fix 10 second blocking delay

bbx10:
usb.Task() has a 10 second blocking delay for all devices that doe IN transfers. This makes it hard to use USB host with other devices that need regular polling. For example, if data is streaming in to Serial1 at 115200 bits/s, the incoming data will not be processed for up to 10 seconds.

This can be seen by adding timing output to the keyboard example. The following prints "delta millis 10000" when there is no keyboard input.

```c
void loop() {
  uint32_t delta, startMillis = millis();

  // Process USB tasks
  usb.Task();

  delta = millis() - startMillis;
  if (delta > 10) {
    Serial.print("delta millis ");
    Serial.println(delta);
  }
}
```

### DIFF
--- a/libraries/USBHost/src/Usb.cpp
+++ b/libraries/USBHost/src/Usb.cpp
@@ -263,6 +263,7 @@ uint32_t USBHost::InTransfer(EpInfo *pep, uint32_t nak_limit, uint8_t *nbytesptr
             continue;
 		}
         if(rcode) {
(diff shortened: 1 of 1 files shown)


ladyada:
just tested this...made keyboardcontroller waaay more responsive. thx!

cmaglie:
Rebased and merged. Thank you!

PR #90, opened 2016-01-12
Fixed attachInterrupt() for use during sleep

rocketscream:
The current implementation of attachInterrupt() will not work on Arduino Zero during sleep (standby mode). By default, all the pins wake up capability during sleep is disabled (Section 20.8.9 Wake-Up Enable in SAMD21 datasheet). Code is tested on a Arduino Zero for pins 0 - 24 (except pin 4). Without these changes, the processor will not wake up if it depends solely on the pin state change as the only source of waking up.

### DIFF
--- a/cores/arduino/WInterrupts.c
+++ b/cores/arduino/WInterrupts.c
@@ -66,6 +66,9 @@ void attachInterrupt(uint32_t pin, voidFuncPtr callback, uint32_t mode)
     enabled = 1;
   }
 
+  // Enable wakeup capability on pin in case being used during sleep
+  EIC->WAKEUP.reg |= (1 << in);
+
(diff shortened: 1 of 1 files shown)


drewfish:
nice

cmaglie:
@ArduinoBot build this please

sandeepmistry:
@rocketscream thanks!

eriktheV-king:
thanks, I'll try it very soon !

zstergios:
Does it work with BLUE PILL SMT32?

maxholgasson:
Hello, does it -external interupt during sleep- work with your Mini Ultra Pro V2 (With Radio)?

PR #92, opened 2016-01-15
[USB CDC] Only re-enable IRQ if PRIMASK was 0 before disabling IRQ

sandeepmistry:
As discussed by @matthijskooijman in https://github.com/arduino/ArduinoCore-samd/pull/65#issuecomment-171951566:

> Also, it might be good to let accept() use the same conditional interrupts() call (in a separate commit, of course)?

Also, incorporated feedback from https://github.com/arduino/ArduinoCore-samd/pull/65#issuecomment-171697817

> To emphasize the non-portability (and reduce confusion for people reading the code), it might be better to even just use __disable_irq() and __enable_irq() directly?


matthijskooijman:
Looks good! I would remove the { } from the if for conciseness, but that's a matter of style/taste :-)

cmaglie:
Rebased and merged.

PR #97, opened 2016-01-19
Added externs for additional Wire instances

spiderkeys:
When working with our SAMD21J18A variant, we implement three I2C interfaces, and I found that the extern declarations were missing when I tried to use Wire1 and Wire2. Added support here for all six potential I2C instances.

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -107,5 +107,10 @@ class TwoWire : public Stream
 };
 
 extern TwoWire Wire;
+extern TwoWire Wire1;
+extern TwoWire Wire2;
+extern TwoWire Wire3;
+extern TwoWire Wire4;
+extern TwoWire Wire5;
 
 #endif
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -106,11 +106,23 @@ class TwoWire : public Stream
     //static const uint32_t XMIT_TIMEOUT = 100000;
 };
 
-extern TwoWire Wire;
-extern TwoWire Wire1;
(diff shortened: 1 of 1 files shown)


drewfish:
Do `Serial` and `SPI` take a similar approach?

spiderkeys:
SPI does, although I see it utilizes the same #ifdef conditionals as the .cpp files for both SPI and the Wire library when it comes to actually creating the instance:

```
#if SPI_INTERFACES_COUNT > 0
  extern SPIClass SPI;
#endif
#if SPI_INTERFACES_COUNT > 1
  extern SPIClass SPI1;
#endif
etc...
```
There is no harm adding that here for the Wire library as well, so I'll do that.

As for the Serial class (Uart), it doesn't manifest itself as an external library and exists as part of the board core. The externs for the Uart objects live in variant.h and the instancing happens in variant.cpp. Leads me to wonder if Serial should be brought out and treated as a library the same way as SPI and Wire. I wouldn't argue for bringing Wire and SPI into the core. Either way, the use of the SERCOM instances is currently somewhat inconsistent.

sandeepmistry:
@spiderkeys do you have a link to your "SAMD21J18A variant"?

spiderkeys:
Still a work in progress, but it currently produces working sketches with all peripherals verified except SPI:
https://github.com/OpenROV/OROV-ArduinoCores

It has some very hacky recipes with hardcoded paths, because we build with arduino-builder on an ARM computer, which currently isn't automatically detecting paths for runtime tools and libs correctly.

sandeepmistry:
@spiderkeys thanks!

Issue #94, opened 2016-01-19
digitalWrite() unnecessarily activates the pull-up resistor on the specified pin


BlackBrix:
digitalWrite() unnecessarily activates the pull-up resistor on the specified pin.

the issue with the pull-up in digitalWrite(), 
(other than the fact that this line shouldn't be there),
is because the pin configuration register (PINCFG) isn't logically ORed with the bit mask when the pull-up bit (PULLEN) is enabled, 
instead the mask is copied directly into the register. 
This effectively clears all the other bits in the register including the input enable (INEN).

in addition this issue makes the code not forward/backward compatible to AVR code
(digitalRead therefore cannot read back the current value of a digital PIN configured as OUTPUT.)

--> https://forum.arduino.cc/index.php?topic=367517.msg2534096#msg2534096
--> http://forum.arduino.cc/index.php?topic=334073.msg2306580#msg2306580

spiderkeys:
I also commented out that line in digitalWrite(), since our custom board has its own pull-up resistors and it was just unnecessarily pulling down the current on many of the lines.

```c++
  PIO_INPUT,            /* The pin is controlled by PORT and is an input. */
  PIO_INPUT_PULLUP,     /* The pin is controlled by PORT and is an input with internal pull-up resistor enabled. */
  PIO_OUTPUT,           /* The pin is controlled by PORT and is an output. */
```
I think it might be prudent to add one more PIO type here, PIO_OUTPUT_PULLUP, and have the functions in the wiring files respect that, or add a separate property in the pin description structure for resistor configuration altogether.

Nothing critical, but it had me scratching my head for a bit, wondering why my LED was barely turning on.

sandeepmistry:
@Dirk67 the goal of enabling the pull-up was to maintain compatibility with AVR core behaviour:

The following is equivalent to ```pinMode(pin, INPUT_PULLUP);``` on AVR:
```arduino
pinMode(pin, INPUT);
digitalWrite(pin, HIGH);
```

However, as you mentioned above, the core is currently clears input enable, so this does not work.

@Dirk67 @spiderkeys I've submitted #101 to resolve this, please review and/or test it out if you have time.

sandeepmistry:
Closing since #101 is merged now.

PR #41, opened 2015-09-21
[samd] fixing issue #28

aethaniel:
This patch should fix https://github.com/arduino/ArduinoCore-samd/issues/28.

Extensive tests needed before merge, just in case of.

PR #101, opened 2016-01-21
Enable digitalRead on output pins, only enable pull-up in digitalWrite if pin is not output

### LINKED ISSUE #56: writing LOW to a digital input pin blocks subsequent digitalRead attempts
I noticed while trying to add Arduino Zero support for Firmata that the following line fails: https://github.com/firmata/arduino/blob/master/examples/StandardFirmata/StandardFirmata.ino#L266.
After that point, digital input will not work on that pin (it's not permanently broken as in fried input pin or anything like that though). This has never been an issue for other architectures.

One one hand this points out that I need to update the Firmata implementation to use `INPUT_PULLUP`, something I should have probably added long ago. On the other hand this is an inconsistency so just calling that out.

This may be the same issue as #28.

sandeepmistry:
Resolves #56 and #94.

* ```pinMode```:
 * enable input mode of output pins, to support using ```digitalRead``` to read current value (maintains AVR behaviour).
 * disable pull-up on output mode
* ```digitalWrite```:
 * only enable pull-up if pin is not in output mode and ```HIGH``` is passed in, disable pull-up if ```LOW``` passed in. (This is similar to #41, but it did not set OUT which is needed to enable pull-up - as per table 22-1 in the SAMD21 datasheet).
 * simplify some duplicated code, by using local variables.

If this is merged #41 can be closed.

### DIFF
--- a/cores/arduino/wiring_digital.c
+++ b/cores/arduino/wiring_digital.c
@@ -88,7 +88,7 @@ void digitalWrite( uint32_t ulPin, uint32_t ulVal )
 
   if ( (PORT->Group[port].DIRSET.reg & pinMask) == 0 ) {
     // the pin is not an output, disable pull-up if val is LOW, otherwise enable pull-up
-    PORT->Group[port].PINCFG[pin].bit.PULLEN = (ulVal != LOW) ;
+    PORT->Group[port].PINCFG[pin].bit.PULLEN = ((ulVal == LOW) ? 0 : 1) ;
   }
 
   switch ( ulVal )
--- a/cores/arduino/wiring_digital.c
+++ b/cores/arduino/wiring_digital.c
@@ -82,8 +82,10 @@ void digitalWrite( uint32_t ulPin, uint32_t ulVal )
     return ;
   }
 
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@ArduinoBot build this please

Dirk67:
digitalWrite: if the pin is in OUTPUT mode, please do not enable/disable/modify the pull-up at all

will it be this way ?

sandeepmistry:
@Dirk67 yes, this is on of the goals of this pull request, in particular see: https://github.com/sandeepmistry/ArduinoCore-samd/blob/digital-read-write-pullup/cores/arduino/wiring_digital.c#L89-L92

Dirk67:
is it within the nightly build already ?

sandeepmistry:
@ArduinoBot build this please

sandeepmistry:
@Dirk67 this change has not been merged into master yet, see @ArduinoBot's comment above for instructions on how to try out these changes.

BlackBrix:
(I am formerly @Dirk67)
I tested this and get no errors.
I think this should merged into master now.

PR #102, opened 2016-01-26
[USB Device] sendStringDescriptor fixes

sandeepmistry:
Based off @cmaglie's [usb-send-descriptor-fix branch](https://github.com/cmaglie/ArduinoCore-samd/tree/usb-send-descriptor-fix) with a few modifications.

Tested with OS X 10.11.3, I can now see "Arduino Zero" in the USB section of the OS X System Reporter app.

Alternative to #75.

@agdl would you be able to test this out on Windows and Linux?


sandeepmistry:
@ArduinoBot build this please

PR #111, opened 2016-02-11
Force link _printf_float in dtostrf

### LINKED ISSUE #109: String() class fails to construct properly with floats and doubles
Hitting an issue where using Arduino's String class constructor fails to create a string when given an initial value that is a float or a double.

Simple repro: Serial.print(String(1.1)) doesn't print anything.

A repro program is attached. This program works when targeting an Uno. This is with the following two packages installed:
◾Arduino SAMD Boards (32-bits ARM Cortex-M0+), version 1.6.2
◾Arduino SAMD Core - (build 1), version 1.6.2-build-2

[string_bug_repro.txt](https://github.com/arduino/ArduinoCore-samd/files/125370/string_bug_repro.txt)

sandeepmistry:
Resolves #109.

```_printf_float``` will only get linked in when ```dtostrf``` is used.

Alternative is to add a ```-u _printf_float``` linker flag, but this will increase sketch size for all sketches even if float support for printf is not needed.
(https://github.com/32bitmicro/newlib-nano-1.0/blob/master/newlib/README.nano)

### DIFF
--- a/cores/arduino/avr/dtostrf.c
+++ b/cores/arduino/avr/dtostrf.c
@@ -20,6 +20,8 @@
 #include <stdio.h>
 
 char *dtostrf (double val, signed char width, unsigned char prec, char *sout) {
+  asm(".global _printf_float");
+
   char fmt[20];
   sprintf(fmt, "%%%d.%df", width, prec);
   sprintf(sout, fmt, val);


sandeepmistry:
@ArduinoBot build this please

cmaglie:
@ArduinoBot build this please

PR #112, opened 2016-02-12
Add support to link the ARM math (provided by CMSIS tools)

sandeepmistry:
Previously, CMSIS was as in the include path, so `arm_math.h` could be included, but there were linker errors when you actually tried to use the functions.

``` c
#define ARM_MATH_CM0PLUS
#include <arm_math.h>
```


sandeepmistry:
@cmaglie @agdl I've rebased this, can one of you please try it out?

Sketches sizes are the same if ARM math is not used in my testing.

PR #113, opened 2016-02-13
Add ASCII control codes

bbx10:
0x00 CTRL-@ through 0x1F CTRL-_

Also return ASCII codes for keys: ESCAPE, DELETE, TAB, and BACKSPACE keys.
Return correct code for ENTER and keypad ENTER keys.


cmaglie:
Rebased and merged. Thank you!

PR #116, opened 2016-02-23
USB CDC ZLP handling

sandeepmistry:
Fixes the handling of ZLP's conditions and USB CDC.

That are experienced when using the native port with the WiFi101 Firmware update tool and OS X.


sandeepmistry:
Closing this in favour of #159.

PR #117, opened 2016-02-23
Make attachInterrupt work as documented

sandeepmistry:
From the [```attachInterrupt``` docs](https://www.arduino.cc/en/Reference/AttachInterrupt), the recommended usage is:

```arduino
attachInterrupt(digitalPinToInterrupt(pin), ISR, mode);
```

This currently does not work with the SAMD core because ```digitalPinToInterrupt ``` is used inside ```attachInterrupt```, so the pin number must be passed in.

This patch addresses this by changing ```digitalPinToInterrupt ``` to return the argument passed in. ```attachInterrupt``` and ```detachInterrupt``` will use ```g_APinDescription[pin].ulExtInt;``` to map the pin to interrupt number directly.

It also moves ```digitalPinToInterrupt``` inside ```Arduino.h``` (if ```ARDUINO_SAMD_VARIANT_COMPLIANCE``` is >= 1.6.5). Existing behaviour will be used in variants with older compliance levels.

I've also added a new change log for variant compliance changes.

### DIFF
--- a/VARIANT_COMPLIANCE_CHANGELOG
+++ b/VARIANT_COMPLIANCE_CHANGELOG
@@ -1,0 +1,8 @@
+SAMD CORE 1.6.5
+
+* digitalPinToInterrupt #define moved to Arduino.h, variant.h must no longer define it
+
+SAMD CORE 1.6.3
+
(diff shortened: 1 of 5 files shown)


sandeepmistry:
@cmaglie I've rebased this, bumped compliance level to 1.6.6 (since 1.6.5 is released) and also applied to the MKR1000 variant. Please review.

PR #122, opened 2016-03-08
Bootloader LED pulsing and TX/RX blinking

### LINKED ISSUE #1: bootloader - TX led not blink when uploading
please add TX, RX led flashing on upload to bootloader

sandeepmistry:
Resolves #1.

Builds upon #64 from @aethaniel and #69 from @facchinm.

Uses SysTick for RX/TX LED pulsing so RX LED is not lit like in #64.


cmaglie:
Superseded by #165

PR #125, opened 2016-03-20
Upstreammerge 150304

ladyada:
tested!


ladyada:
gah that was not the button i meant to press

PR #128, opened 2016-04-04
recipe.ar.pattern backwards compatibility

sandeepmistry:
Port of https://github.com/arduino/Arduino/pull/4737

cc/ @per1234


cmaglie:
Thanks @per1234 !

PR #151, opened 2016-06-30
Fixed PWM frequency glitch

cmaglie:
Previously the TC timer used for PWM was disabled and reenabled on each call to `analogWrite`, this lead to inconsistent PWM periods.

This PR fixes this behaviour.


agdl:
@cmaglie good since previously holes were present in the waveform. We should also change the PWM frequency to be consistent with other boards.
Now PWM frequency on SAMD is ~186kHz, on AVR is ~960Hz

cmaglie:
At the moment is 48MHz / 256 = 187.5KHz, we can prescale by another 256 and get a (weird but) more manageable: 187.500 / 256 = 732.4Hz

PR #163, opened 2016-09-14
Increase buffer size for converting floats + doubles to String's

sandeepmistry:
Previously it was too small to print FLT_MAX and DBL_MAX.

Troublesome sketch:

``` arduino
#include <float.h> 

void setup() {
  Serial.begin(9600);
  while (!Serial);

  Serial.println(String(FLT_MAX));
  Serial.println(String(DBL_MAX));
  Serial.println(String(-FLT_MAX));
  Serial.println(String(-DBL_MAX));
}

void loop() {
}
```

### DIFF
--- a/cores/arduino/WString.cpp
+++ b/cores/arduino/WString.cpp
@@ -19,6 +19,8 @@
   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#include <float.h>
+
 #include "WString.h"
 #include "itoa.h"
 #include "avr/dtostrf.h"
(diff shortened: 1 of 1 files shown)


sandeepmistry:
If this is merged, it also needs to be ported to the 101 core.

Ivan-Perez:
Is that code correct? You shouldn't declare an array whose size is not available at compilation time.

I can't test it now, but if you enable all warnings the compiler should emit one because of it. Any way, as I'm reading [here](http://stackoverflow.com/a/15013295/1852787), g++ may allow it as an extension, but it is not valid C++.

aentinger:
Closing this because it will be done in https://github.com/arduino/ArduinoCore-API/issues/128 instead.

PR #104, opened 2016-01-29
[Bootloader] Enable USB string descriptors

sandeepmistry:
Add support for sending USB string descriptors for manufacturer and product in the boot loader.

The following error is no longer present in OS X 10.11.3 kernel logs, after these boot loader changes are applied.

```
2016-01-25 4:33:08.000 PM kernel[0]: 129020.020709 IOUSBHostDevice@14200000: IOUSBHostDevice::getDescriptorGated: type 0x03 index 0 length 2 completed with 0xe0005000 and bytesTransferred 0
2016-01-25 4:33:08.000 PM kernel[0]: 129020.022294 IOUSBHostDevice@14200000: IOUSBHostDevice::getDescriptorGated: type 0x03 index 0 length 2 completed with 0xe0005000 and bytesTransferred 0
```

Tested with OS X 10.11.3, and Linux (Rasbian Jessie).

cc/ @aethaniel

PR #165, opened 2016-09-22
Bootloader LEDs / TXRX activity / USB Descriptors / Makefile

### LINKED ISSUE #1: bootloader - TX led not blink when uploading

cmaglie:
This is a rebase and merge of:

#122 Bootloader LED pulsing and TX/RX blinking
#104 Enable USB string descriptors
#69 Add Leonardo-like led pulse when bootloader is active
#64 Bootloader leds activity

Moreover:
- each one of the PR above added a feature that is board-specific (LEDs pins, USB descriptors, etc.) so instead of adding more variables to pass to compiler I've changed the build process so each board can be defined through his own `board_descriptions_xxxx.h` file.
- the TX/RX leds and on-board LED can now be disabled
- I've added a script to build all the bootloaders altogether

Tested with Arduino Zero and MKR1000 on Linux.

Fix #1 

/cc @aethaniel @facchinm @sandeepmistry


sandeepmistry:
Looks good on OS X 10.11.6 + IDE 1.6.12 + Arduino Zero (Native USB)

* LED pulses in boot loader mode
* double tap via USB puts board in boot loader mode
* RX and TX LED's blink on sketch upload
* OS X shows USB string descriptors

screen shot 2016-09-22 at 2 14 04 pm

sandeepmistry:
Looks good on my Windows 10 VM as well. I think we should still test with a non-VM windows PC though.

PR #170, opened 2016-10-03
Fixed issue where pins were not being configured properly in analogWr…

bose-mdellisanti:
…ite when multiple channels of the same TCC/TC module were being used to drive outputs.

The recent release of 1.6.7 introduced a bug that prevented pins from being configured properly in PWM mode.  This patch just rearranged a small bit of code to make sure that this doesn't occur.

The original commit where the bug was created was a few months back.  I posted a little more info over on that commit:
https://github.com/arduino/ArduinoCore-samd/commit/f221d1f1bdb3bfd87ca3f20520f67bef9450619e#diff-63b7293bbdf8f8f0f8b98336a5e85d83


cmaglie:
Thank you @bose-mdellisanti !

Issue #169, opened 2016-09-30
HX711 library (Queuetue) no longer talks to device with 1.6.7


totalgee:
Hello there, I have a program that's talking to a HX711 (using the [Queuetue library](https://github.com/queuetue/Q2-HX711-Arduino-Library)), and it worked fine with v1.6.6 of the SAMD tools (I'm using MKR1000).  Today I updated to v1.6.7, and the same program, same hardware, no longer seems to talk to the HX711, it just always returns zero as the data measurement.

Any tips on what might have changed that could cause this?  The HX711 is hooked up with data on A2 and clock on A3, and there is another device using SCL/SDA for I2C (that device is working with both versions of the MKR1000 driver and tools).

If I revert back to 1.6.6, everything works fine again.  Thanks for any thoughts on how to solve this.

sandeepmistry:
Hi @totalgee,

I've tested `shiftIn`, `digitalWrite` and `digitalRead` with 1.6.7, things seem ok to me. I don't have a HX711 to test with unfortunately.

There have been some changes to `pinMode` and `digitalWrite` in #101, which disable the pullup in output mode. Not sure if that is the root cause of you issues.

Could you try replacing `cores/arduino/wiring_digital.c` with [the version from 1.6.6](https://github.com/arduino/ArduinoCore-samd/blob/1.6.6/cores/arduino/wiring_digital.c)? To see if that was the relevant change. Also, does the same thing happen if you change pins?

totalgee:
Hi, so I tried replacing the wiring_digital.c with the old version from 1.6.6, and it doesn't seem to fix the problem when I rebuild and upload my MKR1000 project.

Just to confirm, do I need to do anything special to build it?  I found the file in `~/Library/Arduino15/packages/arduino/hardware/samd/1.6.7/cores/arduino` (I'm on OS X) and replaced it with the older file.  When I build my project, how do I confirm it's actually compiled and using the modified file?  Are there any other steps to perform to test this properly?  Thanks.

totalgee:
If it helps, the `hx711.readyToSend()` function never returns true any more.  The [code in the library](https://github.com/queuetue/Q2-HX711-Arduino-Library/blob/master/src/Q2HX711.cpp) for this function is:

```c++
bool Q2HX711::readyToSend() {
  return digitalRead(OUT_PIN) == LOW;
}
```

[...]

PR #174, opened 2016-10-11
Launch libc initialization after hardware setup

cmaglie:
This pachs allows C++ global constructors to run after hardware initialization. This helps some libraries that setups hardware in class constructor to work properly.

See also #169

/cc @totalgee


totalgee:
Wonderful, thanks!  I just tried with this code locally on top of 1.6.7 with the original version of Q2HX711 (which calls `pinMode()` to setup an OUTPUT in a global constructor) and it does indeed work now. (ref #169)

PR #179, opened 2016-10-26
PWM proposal

agdl:
This commit enables PWM on pin 11 instead of pin A4 on MKR1000 improving UNO pinout compatibility


sandeepmistry:
I've slightly modified this in https://github.com/arduino/ArduinoCore-samd/commit/166df7c096e019fa62657786b10c95c971920e87. Now A4 and 11 share the timer.

@agdl please make a note to update the product documentation page once the next SAMD core is released: https://www.arduino.cc/en/Main/ArduinoMKR1000 (cc @akash73)

PR #180, opened 2016-11-07
Optimize SPI transfers

sandeepmistry:
Combines `SERCOM::writeDataSPI` and `SERCOM::readDataSPI()` into a single `SERCOM::transferDataSPI` method and remove some flags checks.

Needs more testing, so far basic tests with WiFi101 and SD libraries seem ok.

cc/ @cmaglie @agdl

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -300,24 +300,11 @@ void SERCOM::setClockModeSPI(SercomSpiClockMode clockMode)
   enableSPI();
 }
 
-void SERCOM::writeDataSPI(uint8_t data)
+uint8_t SERCOM::transferDataSPI(uint8_t data)
 {
-  while( sercom->SPI.INTFLAG.bit.DRE == 0 )
-  {
-    // Waiting Data Registry Empty
-  }
-
   sercom->SPI.DATA.bit.DATA = data; // Writing data into Data register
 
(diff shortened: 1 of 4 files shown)


dlabun:
Is it possible to take the optimizations one (major) step further and introduce DMA?

sandeepmistry:
@dlabun 

> Is it possible to take the optimizations one (major) step further and introduce DMA?

It's something to look into later, however due to the blocking nature of the current `SPI.transfer(...)` API's, I'm not expecting it to yield any performance gains. Also, from testing with the `SD` and `WiFi101` libraries, almost all of the SPI transfers are byte by byte.

AloyseTech:
Hi guys, I'm working on a DMA Helper to create dedicated DMA library (SPI, I2C, USART...). I have the SPI with DMA working. I still have issue : I can't create an DMA_SPI object in other classes. I think it is a scope issue but I don't have enough knowledge to correctly deal with this issue. ANy help would be appreciated !
Take a look at the source and Unit test here : [https://github.com/aloysetech/SAMD_DMA_Helper](https://github.com/aloysetech/SAMD_DMA_Helper)

dlabun:
@sandeepmistry 
I certainly understand, right now I don't see an issues with writing to my SD card. I do see performance issues with the LCD and that's where I think DMA would help the community.

@AloyseTech 
I'm sure I know even less about coding than you, but would certainly test out your library on the Sharp memory LCD once the library is functional.

PR #182, opened 2016-11-17
Add I2S support

sandeepmistry:
WIP, for review and testing only for now ...

cc/ @cmaglie @facchinm @agdl

### DIFF
--- a/cores/arduino/wiring.c
+++ b/cores/arduino/wiring.c
@@ -65,6 +65,7 @@ void init( void )
     // Capture error
     while ( 1 ) ;
   }
+  NVIC_SetPriority (SysTick_IRQn,  (1 << __NVIC_PRIO_BITS) - 2);  /* set Priority for Systick Interrupt (2nd lowest) */
 
   // Clock PORT for Digital I/O
 //	PM->APBBMASK.reg |= PM_APBBMASK_PORT ;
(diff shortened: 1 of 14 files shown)


agdl:
Tested and LGTM!

colintd:
Hi,  I'm doing some detailed timing work on a SAMD21 board, and think this check-in might have broken the micros() function.

To allow micros() to be called from an ISR, the micros() function checks the pending flag on the systick handler.  If this is set, it assumes the counter has wrapped but the tickcounter hasn't yet been incremented,, and adds 1ms to the result.  This was fine when the handler was the highest priority, as it was either pended or it ran to completion.  With this change to reduce the priority, there is a window where the handler is called, but then preempted by a higher priority external interrupt (such as an EIC interrupt from a GPIO pin).  In that case, the tick count may or may not have been incremented (we can't tell), which means some of the time we add an extra 1ms when we shouldn't.

I've spotted  this problem as I'm using micros() to timestamp edges for which I know the timing (1 pulse per second). I see multiple stamps an hour with exactly 1ms error due to this bug.

Can you elaborate why this change is needed for I2S support?

matthijskooijman:
@sandeepmistry, could you have a look at the above comment? IMHO, something like changing the systick interrupt priority should at least have had a comment in the code and commit message, maybe even have been a separate commit for clarity.

PR #184, opened 2016-12-06
Add support for different bootloader sizes

awatterott:
Currently the bootloader size is hard coded to 8k, because APP_START is not modifiable.
Can we change this?

### DIFF
--- a/cores/arduino/Reset.cpp
+++ b/cores/arduino/Reset.cpp
@@ -24,7 +24,9 @@ extern "C" {
 #endif
 
 #define NVM_MEMORY ((volatile uint16_t *)0x000000)
-#define APP_START 0x00002004
+#ifndef APP_START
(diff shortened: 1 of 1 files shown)


facchinm:
Hi @awatterott ,
you are right, the `APP_START` value is fixed and it shouldn't, but it only handles the `banzai` function, which is called in case of 1200bps reset. You can already use a different bootloader size and things work just good (if you don't specify `1200_bps_touch` on the `boards.txt`).
Anyway, it should be fixed but in a different way I believe... I'll post a proposal here really soon

awatterott:
Hi,
thanks for the fast reply. We are working on a new bootloader and we also want to use the 1200bps baudrate change to start it.

facchinm:
Fixed by #185

PR #185, opened 2016-12-07
Add method to automatically retrieve sketch start

facchinm:
Superseded #184

If `ARDUINO_SAMD_VARIANT_COMPLIANCE` is >= 1.6.10 you should define ` __text_start__ ` symbol in the linker script (as in the examples). This allows `APP_START` variable to automatically acquire the right position.

Another possible improvement is checking if APP_START == 0, which means that `without_bootloader` linker script was used. In this case, calling `banzai` should be avoided and a simple restart should be performed.

@awatterott @sandeepmistry @cmaglie

### DIFF
--- a/VARIANT_COMPLIANCE_CHANGELOG
+++ b/VARIANT_COMPLIANCE_CHANGELOG
@@ -1,3 +1,7 @@
+SAMD CORE 1.6.10
+
+* the linker script is use should define __text_start__ symbol at the beginning of .text section
+
 SAMD CORE 1.6.6
 
 * digitalPinToInterrupt #define moved to Arduino.h, variant.h must no longer define it
--- a/cores/arduino/Reset.cpp
+++ b/cores/arduino/Reset.cpp
@@ -24,7 +24,15 @@ extern "C" {
 #endif
 
 #define NVM_MEMORY ((volatile uint16_t *)0x000000)
+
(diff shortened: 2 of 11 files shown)


sandeepmistry:
Don't forget to update `VARIANT_COMPLIANCE_CHANGELOG` if we decide to merge this.

Also, if the linker scripts are the same for all the variants maybe we consolidate them later?

awatterott:
@facchinm Thanks. I will test it at the weekend.
BOSSA will also need some changes to work with different bootloader sizes: https://github.com/shumatech/BOSSA/issues/33
As workaround for BOSSA it is possible to check if the first write address is smaller than the bootloader size and then use it as an offset. 
```
if first_write_address < bootloader_size
  offset = bootloader_size - first_write_address

real_write_address = write_address + offset
```

awatterott:
@facchinm I have tested the changes and everything works.
We will also release our bootloader (sam-ba + mass-storage) in the next days: https://github.com/watterott/SAM-BAR

tuxedo0801:
@awatterott 
this might be offtopic, but: is there already some info about your upcoming bootloader already online?!

awatterott:
@tuxedo0801 It is a SAM-BA compatible bootloader with an integrated 2nd mass-storage bootloader and it fits in 16k.

tuxedo0801:
Sounds great. Looking forward for more details.

PR #190, opened 2016-12-16
Integrate Arduino.org boards

facchinm:
Tian needs https://github.com/arduino/Arduino/pull/5731/commits/e9709689de95cf3e945c6a1dfe8e113fab64e97e to work correctly. 
Avrdude needs to be added as a dependency in index json file


dlabun:
Question: Is it really wise to integrate the Tian in with the rest of the SAMD21 core and not give it a separate core? It seems like there is alot of additional code is being added to support a single board and that code doesn't apply any of the official or 3rd SAMD21 based boards.

facchinm:
Hi @dlabun ,
in fact the Tian share its variant with all the other Arduino.org boards, so it "naturally" lives in the same repo. The only particular feature is a 2 ports serial-to-usb converter which exposes with the same VID/PID pair on both ports. Only one of those can be used for uploading, the other one is a console directly connected to the Linux processor. Once this is solved, its maintenance will be almost zero :smile:

facchinm:
Merged manually

PR #192, opened 2016-12-19
Update boot loader source to use new CMSIS packages

sandeepmistry:
cc/ @cmaglie @facchinm


facchinm:
Am I missing something or the conversions like 
`pUsb->DEVICE.DeviceEndpoint[0].EPINTFLAG.bit.* = true` and `pUsb->DEVICE.DeviceEndpoint[0].EPINTFLAG.reg = USB_DEVICE_EPINTFLAG_*`
are clearing all the register data and thus changing the expected behaviour?

cmaglie:
Right, for `EPINTFLAG` the `.bit` or bitwise-or `|=` notation is fine and should not be changed.

The change above must be made for `EPINTENCLR` to avoid read-change-store operation on a register that is clear-on-set.

cmaglie:
Well reading again the datasheet `EPINTFLAG` actually *is* a clear on write register:

```
Bit 6 – STALL1: Transmit Stall 1 Interrupt Flag

This flag is cleared by writing a one to the flag.
....
Writing a zero to this bit has no effect.
Writing a one to this bit clears the STALL1 Interrupt Flag.
```

and the same happens for the other bits in the register, so for this part the PR is correct. :+1:

PR #195, opened 2017-01-01
Fix utoa type conflict

kbumsik:
This is corresponding to [an issue in the main Arduino repo](https://github.com/arduino/Arduino/pull/5480).
The type conflict problem happened to me when using the new newlib/arm-gcc with Atmel Studio 7.
The fix also works with Arduino version 1.6.11.


sandeepmistry:
@kbumsik thank you!

Issue #200, opened 2017-01-08
External interrupts take 6 us to responds


joverbee:
Even though we can expect the interrupt handling through arduino interface to be slower than doing low level coding with the dataheet at hand, 6 us is really ridicoulously long for a processor with 20ns clock cycle (300 clock cycles).
I havent figured out yet to find the cause of this (played with genclck of the EIC but no improvement) but to me it is clear that this can not be an optimal way to use the EIC controller. This makes interrupting for anything dynamic impossible. It seems even longer than it was on arduino uno boards, but dont recall the exact time.
( btw I managed through hw event system to start stop the TCC on external interrupt in 180ns=9 cycles- but of course we need to do more checking and jumping to ISR in this case, but still: does it really take 291 clock cycles to do this?)

joverbee:
i checked the code and the culprit is the EIC handler which checks all 16 interrupt channels even when they are not enabled. As I was working with extint14 this means my ISR is preced by 14 reads of the status register which is a total waste of time.
I plan to rewrite the code in Winterrupts.c to make a stack of attached interupts with their masks for checking the status bit. i would then service the isrs on a first come first serve basis. This should dramatically speed up interrupts especially when not all 16 channels are in use ( that would be a very rare case anyway)

Can anyone point me to a simple description on how I have to tell the IDE to work with my version of the core rather than the one supplied ? THis would really allow me to focus on this issue rather than on figuring out how to make it compile.

dlabun:
The quick and dirt method is to just modify the copy of the core installed on your computer... The only draw back is if you update the core through the Board Manager you wipe out all over your modifications.

joverbee:
That is what I figured out doing. But I also made a pull request, so hopefully the next version that comes in through the board manager will contain my changes and we all profit ;)

cmaglie:
Fixed by #261

Issue #199, opened 2017-01-07
Changing trigger level in attach interrupt during runtime


joverbee:
Current behaviour in Winterrupts.c assumes we perform attachinterrupt(pin,ISR,mode) only once after reset. However if you want to dynamically change the trigger mode, this leads to wrong behaviour:
e.g first starting with RISING and then FALLING leads actually to BOTH because the mode bits are ORed into the register. 

I assume putting this in front of the switch(mode) in attachinterrupt solves the issue
```
EIC->CONFIG[config].reg &=~ (EIC_CONFIG_SENSE0_Msk << pos); //reset sense mode
  switch (mode) 
```
However I totally fail to explain my setup that it should recompile this file and use my own version rather than the default one? Any help would be appreciated...its driving me crazy (I seem to always get stuck for hours on include issues...aaargh)

facchinm:
Hi @joverbee ,
if you modified the right file, it should be automatically recompiled. What do you mean with "I seem to always get stuck for hours on include issues" ? However, the patch looks good, comments on its own thread

joverbee:
I figured it out. I am working on Mac and it wasn't really clear where the actual Winterrupts.c file was that the ide was using (I had a few different one scattered over my drive). I found out by trying to redefine EIC_Handler in the arduino code which then made the compiler complain that he already had a definition in this-that-directory.

cmaglie:
Thanks! Fixed by a9fa694.

PR #201, opened 2017-01-08
solve issue #199, improve issue #200 in Winterrupts

joverbee:
#199 corrected an error in the behaviour of attachinterrupt when changing the trigger mode of an interrupt. I now first erase the relevant trigger mode bits, before OR-writing a new setting. (this erasing was missing, causing unpredictable behaviour when calling attachinterrupt more than once for the same pine and with different trigger modes.

#200 I noticed that the EIC_Handler had to loop over all possible external interrupts and check if an interrupt occurred even if we didn't enable those interrupt. This resulted in up to 6us wait times before e.g. EXTINT14 was serviced. I rewrote the handler to keep a list of all attached interrupts (and a handle to their ISRs) with a first come first serve policy, meaning that the first attached interrupt is treated first. This improves the wait time for a single interrupt with edge triggering to slightly less than 2us. This is still long, but already a clear improvement. My guess is that the jump to an ISR with 32 bit address that is stored in RAM is now the main culprit of the remaining time, but this would probably need some inline assembly or a better grasp of how the compiler turns this code into seemingly inefficient assembly code...but hey a factor of 3 is already something. Remember that speed might be an important reason why people decided to buy the Zero (or variants) in the first place.


facchinm:
The rationale and the patches look good, I'd only make some slight adjustments.
* remove the debug code `getISRcallback()` and `getISRlist()`and relevant variables or enclose them in a `#if 0` or similar
* separate the patches so every patch addresses a single issue

joverbee:
I am quite inexperienced with using Github for collaborative projects so excuse if I am doing this wrong. I reverted the changes and made them in 2 separate commit steps, one solving issue #199 and the other bigger change for improving #200. I removed the debug code which also means Winterrups.h is not changed.

cores/arduino/WInterrupts.c
```diff
@@ -76,41 +81,64 @@ void attachInterrupt(uint32_t pin, voidFuncPtr callback, uint32_t mode)
...
+      //this allow for calling attachInterrupt(pin, NULL, mode), we set up all needed register
+      //but won't service the interrupt, this way we also don't need to check it inside the ISR
+
+    // Store interrupts to service in order of when they were attached
+    // to allow for first come first serve handler
+    uint32_t current=0;
+    //check if we already have this interrupt
+    int id=-1;
+    for (uint32_t i=0; i<nints; i++){
+      if (ISRlist[i]==in) id=in;
```
cmaglie: The condition here should be `id = i` I guess?

cores/arduino/WInterrupts.c
```diff
@@ -132,25 +160,37 @@ void detachInterrupt(uint32_t pin)
   
   // Disable wakeup capability on pin during sleep
   EIC->WAKEUP.reg &= ~(1 << in);
+
+  //and take it out of the ISR list
+  int id=-1; //find its location first
+  for (uint32_t i=0; i<nints; i++){
+      if (ISRlist[i]==in) id=in;
```
cmaglie: Same here it too should be `id=i`

cores/arduino/WInterrupts.c
```diff
@@ -132,25 +160,37 @@ void detachInterrupt(uint32_t pin)
...
+  }
+  if (id==-1) return; //apparently we didn't have it
+  for (uint32_t i=id; i<nints-1; i++){ //and shift the ones above one down
+      ISRlist[i]=ISRlist[i+1];
+      ISRcallback[i]=ISRcallback[i+1];
+  }
+  //and remove the top item
+  ISRlist[nints]=0;
+  ISRcallback[nints]=NULL;
```
cmaglie: Is erasing the list element really needed? IMHO just decreasing `nints` should be enough, new elements added later will overwrite the old value.

cores/arduino/WInterrupts.c
```diff
@@ -132,25 +160,37 @@ void detachInterrupt(uint32_t pin)
...
-      // Clear the interrupt
-      EIC->INTFLAG.reg = 1 << i;
+  //calling the routine directly from -here- takes about 1us
+  //depending on where you are in the list it will take longer
+  for (uint32_t i=0; i<nints; i++) // Loop over all enabled interrupts in the list
+  {  
+    if ((EIC->INTFLAG.reg & 1<<ISRlist[i]) != 0)
+    {  
+      ISRcallback[i]();// Call the callback function if assigned
+      EIC->INTFLAG.reg = 1<<ISRlist[i]; // Clear the interrupt
```
cmaglie: I'm wondering if we can gain something if we store `1 << ISRlist[i]` instead of `ISRlist[i]`.

joverbee:
great! glad somebody is fixing this and apologies for my cumbersome interaction with the github process

regards,
Jo

On 5 Sep 2017, at 14:07, Cristian Maglie <notifications@github.com<mailto:notifications@github.com>> wrote:


@cmaglie commented on this pull request.

Hi @joverbee<https://github.com/joverbee>,
I've added some inline comments, I'm actually working on my local copy, so I'll open another PR with a clean tree that includes this fix.

________________________________

In cores/arduino/WInterrupts.c<https://github.com/arduino/ArduinoCore-samd/pull/201#discussion_r136967788>:

> @@ -76,41 +81,64 @@ void attachInterrupt(uint32_t pin, voidFuncPtr callback, uint32_t mode)
   // Assign pin to EIC
   pinPeripheral(pin, PIO_EXTINT);

-  // Assign callback to interrupt
-  callbacksInt[in] = callback;
+  if (callback){ //only store when there is really an ISR to call
+      //this allow for calling attachInterrupt(pin, NULL, mode), we set up all needed register
+      //but won't service the interrupt, this way we also don't need to check it inside the ISR
+
+    // Store interrupts to service in order of when they were attached
+    // to allow for first come first serve handler
+    uint32_t current=0;
+    //check if we already have this interrupt
+    int id=-1;
+    for (uint32_t i=0; i<nints; i++){
+      if (ISRlist[i]==in) id=in;


The condition here should be id = i I guess?

________________________________

In cores/arduino/WInterrupts.c<https://github.com/arduino/ArduinoCore-samd/pull/201#discussion_r136968085>:

> @@ -132,25 +160,37 @@ void detachInterrupt(uint32_t pin)

   // Disable wakeup capability on pin during sleep
   EIC->WAKEUP.reg &= ~(1 << in);
+
+  //and take it out of the ISR list
+  int id=-1; //find its location first
+  for (uint32_t i=0; i<nints; i++){
+      if (ISRlist[i]==in) id=in;


Same here it too should be id=i

________________________________

In cores/arduino/WInterrupts.c<https://github.com/arduino/ArduinoCore-samd/pull/201#discussion_r136968410>:

> @@ -132,25 +160,37 @@ void detachInterrupt(uint32_t pin)

   // Disable wakeup capability on pin during sleep
   EIC->WAKEUP.reg &= ~(1 << in);
+
+  //and take it out of the ISR list
+  int id=-1; //find its location first
+  for (uint32_t i=0; i<nints; i++){
+      if (ISRlist[i]==in) id=in;
+  }
+  if (id==-1) return; //apparently we didn't have it
+  for (uint32_t i=id; i<nints-1; i++){ //and shift the ones above one down
+      ISRlist[i]=ISRlist[i+1];
+      ISRcallback[i]=ISRcallback[i+1];
+  }
+  //and remove the top item
+  ISRlist[nints]=0;
+  ISRcallback[nints]=NULL;


Is erasing the list element really needed? IMHO just decreasing nints should be enough, new elements added later will overwrite the old value.

________________________________

In cores/arduino/WInterrupts.c<https://github.com/arduino/ArduinoCore-samd/pull/201#discussion_r136968985>:

> -    {
-      // Call the callback function if assigned
-      if (callbacksInt[i]) {
-        callbacksInt[i]();
-      }
-
-      // Clear the interrupt
-      EIC->INTFLAG.reg = 1 << i;
+  //calling the routine directly from -here- takes about 1us
+  //depending on where you are in the list it will take longer
+  for (uint32_t i=0; i<nints; i++) // Loop over all enabled interrupts in the list
+  {
+    if ((EIC->INTFLAG.reg & 1<<ISRlist[i]) != 0)
+    {
+      ISRcallback[i]();// Call the callback function if assigned
+      EIC->INTFLAG.reg = 1<<ISRlist[i]; // Clear the interrupt


I'm wondering if we can gain something if we store 1 << ISRlist[i] instead of ISRlist[i].

cmaglie:
Followups on #261

PR #203, opened 2017-01-16
Add support for updating sketches from an SD card

sandeepmistry:
Via 2nd stage boot code that checks SD card for `UPDATE.bin`.

Also, added tool info to `platform.txt`  to update sketches via HTTP outside the core for later.

cc/ @cmaglie @facchinm @agdl @akash73


tuxedo0801:
Is there some kind of documentation on how to use this? Just copy the update.bin to sd-card and reboot arduino zero? Is this feature now part of the default-bootloader that comes with ArduinoCore-samd?

As SPI Flash is connected the same way as sd card, I guess it will work with SPI flash as well?

variants/arduino_zero/linker_scripts/gcc/flash_with_bootloader.ld
```diff
@@ -67,6 +67,9 @@ SECTIONS
 	{
 		__text_start__ = .;
 
+		KEEP(*(.ota_boot))
+
+		. = ALIGN(256);
```
facchinm: I'd align the .ota_boot section to either 0x2000 or 0x4000 to make sure that the reset vector does not get relocated to something different than 0x6000 if the 2nd stage bootloader becomes slightly smaller

sandeepmistry: I'll change it to `0x2000`.

Maybe we can have a clearer name for this section `second_stage`, `second_stage_boot` or `sketch_boot`?

tuxedo0801:
*ping*

cmaglie:
@tuxedo0801 
you must patient a bit, this is still experimental, we'll provide documentation once ready.

tuxedo0801:
@cmaglie thanks for first feedback.
I don't want the complete documentation. Just let me know if this would be tied to sd cards or if it will be possible with SPI flash as well (or maybe it's up to the user?!).

That would help me continuing my development into the right direction...

cmaglie:
For now it's SD card only.

dlabun:
Out of curiosity, is this going to replace the default boot loader or run in tandem with it?

sandeepmistry:
@dlabun good question! It will run in tandem.

tuxedo0801:
To give you an insight in why I want to use SPI flash instead of SD card:

I'm developing small devices for KNX home automation bus. The devices are totally bus-powered. And the KNX transceiver chip allows max ~50mA to be drawn from KNX bus. So it's very low power.

SD cards consume too much energy to be powered from KNX bus. SPI flash consumes almost nothing, compared to SD card. And even micro-sd card socket is much bigger than a small SPI flash package... Size matters ;-) 

As the SPI flash is not pluggable like SD card, the new sketch is received directly from KNX bus. So updating a device is just a simple click on the control-application on a PC that is also connected to bus. File is transferred over bus and stored into SPI flash.

I guess there are more people out there that wand to read the firmware from somewhere else than a pluggable SD card (get via ethernet, wifi, some radio module, power line, ... whatever and stored on any storage).

I would be glad if you foresee a user-friendly mechanism that allows the sketch-writer to speficy an own storage mechanism to read the update from, that don't requires fiddling with core classes and modifying the samd core.

Beside my whishes: Tumbs up for this new feature... Keep up the good work.

deladriere:
what CS pin are you using ? (mine is pin 10)

I can't have the bootloader to boot from the SD
Here is what I did
1) installed as instructed
2) burned the bootloader with Atmel SAM-ICE programmer
3) compiled and renamed the .bin to UPDATE.bin
4) copied the UPDATE.bin to sd
5) reseted the board

but it stays in booloader mode with the led glowing 
Am i missing something ?

deladriere:
Sorry to insist : I can't figure how to get started
thanks !

deladriere:
Where do you set the CS pin to use ?

agdl:
Tested. LGTM

deladriere:
@agdl How do you set the CS pin ?

tuxedo0801:
@cmaglie 

> you must patient a bit, this is still experimental, we'll provide documentation once ready.

It has been merged into master about three weeks ago. So it's ready now?

Questions so far:

* How do I enable this feature? Just include the "SDU" lib in my sketch? 
* Is it possible to use alternative pins for accessing sd-card? If yes, how?
* Is it somehow foreseen to easily use an alternative "SDUBoot.ino" implementation? (like using an SPI flash chip instead of a SD card)

tuxedo0801:
For anyone searching the same information: #220

PR #207, opened 2017-01-24
Set full PINCFG register on pinMode(pin, OUTPUT)

### LINKED ISSUE #205: analogRead() followed by pinMode(,OUTPUT) leaves pin muxed to ADC
Hi,

The following sequence will leave the pin muxed to the ADC and fail on the digital write:

`analogRead(pin);`
`pinMode(pin,OUTPUT);`
`digitalWrite(pin,HIGH);`

A workaround is to insert a `pinMode(pin,INPUT)` before the `pinMode(pin,OUTPUT)`.

I think the issue is with commit `e431bf7` to `wiring_digital.c`.  It sets the INEN bit without clearing the PMUXEN bit.  It is also leaving any PULLEN and DRVSTR setting in place where as the previous revision was clearing the whole register except INEN.  Not sure if the best fix is to just clear PMUXEN, or to set the whole register with PORT_PINCFG_INEN so everything else is cleared.

sandeepmistry:
This change is a proposal to fix #205.

cc/ @kiptonk

### DIFF
--- a/cores/arduino/wiring_digital.c
+++ b/cores/arduino/wiring_digital.c
@@ -58,11 +58,8 @@ void pinMode( uint32_t ulPin, uint32_t ulMode )
     break ;
 
     case OUTPUT:
-      // enable input, to support reading back values
-      PORT->Group[g_APinDescription[ulPin].ulPort].PINCFG[g_APinDescription[ulPin].ulPin].bit.INEN = 1 ;
-
-      // disable pullups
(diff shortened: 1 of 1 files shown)


kiptonk:
This looks good, thanks!

sandeepmistry:
@kiptonk great, thank you for testing! The change will be part of the next SAMD core release.

Issue #209, opened 2017-02-02
Warning on wiring.c


jwillemsen:
We get a warning using platformio with the sodaq boards. Looks the warning is coming from this repo

I am trying to enable travis-ci support using pipeline for the sodaq boards. We want our code to be warning free so use -Werror as option, but it seems there is a warning in the sodaq one support code.

```
Project has been successfully initialized!

Useful commands:

`platformio run` - process/build project from the current directory

`platformio run --target upload` or `platformio run -t upload` - upload firmware to embedded board

`platformio run --target clean` - clean project (remove compiled files)

`platformio run --help` - additional information

[Thu Feb  2 20:03:34 2017] Processing sodaq_autonomo (platform: atmelsam, board: sodaq_autonomo, framework: arduino)

--------------------------------------------------------------------------------

PackageManager: Installing framework-arduinosam @ ~2.10611.0

Downloading...

Unpacking...

framework-arduinosam @ 2.10611.0 has been successfully installed!

Verbose mode can be enabled via `-v, --verbose` option

Converting SodaqAutonomo-basic.ino

Collected 6 compatible libraries

Looking for dependencies...

Library Dependency Graph

|-- <src>

Compiling .pioenvs/sodaq_autonomo/src/SodaqAutonomo-basic.ino.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduinoVariant/variant.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/IPAddress.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/Print.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/Reset.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/RingBuffer.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/SERCOM.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/Stream.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/Tone.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/USB/CDC.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/USB/PluggableUSB.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/USB/USBCore.o

Compiling .pioenvs/sodaq_autonomo/FrameworkArduino/USB/samd21_host.o

[...]

PR #210, opened 2017-02-03
Fixed unused variable warning, see issue #209

jwillemsen:
* cores/arduino/wiring.c:


jwillemsen:
@facchinm For maintainability it looks better to me to move the ul variable to the loop, independent of the fact that this seems to not fix the warning I have.

PR #231, opened 2017-05-01
USB: Only swap incoming buffers for non-ZLP's

sandeepmistry:
I noticed `Serial.available()` was stuck returning `0` when I sent 1033 bytes from the Serial Monitor in the IDE from macOS. Sketch snipping:

```arduino
  do {
    if (SerialUSB.available()) {
      char c = SerialUSB.read();

       SerialUSB.print(c);

       publicCert += (char)c;
    }
  } while (!publicCert.endsWith("-----END CERTIFICATE-----"));
```

It turns out a ZLP was being received, and the USB code was swapping incoming buffers, when it didn't need to and this confused the code in `available()` which never swaps back to the other buffer.

This change prevents the swap in the USB endpoint ISR handler when a ZLP is received. NOTE: the call to `release();` is needed, this is why I didn't use the `if (last0 == 0) { return; }` style.

### DIFF
--- a/cores/arduino/USB/SAMD21_USBDevice.h
+++ b/cores/arduino/USB/SAMD21_USBDevice.h
@@ -319,30 +319,34 @@ public:
 			usbd.epBank0AckTransferComplete(ep);
 			//usbd.epBank0AckTransferFailed(ep); // XXX
 
-			// Update counters and swap banks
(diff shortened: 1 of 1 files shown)


sandeepmistry:
@cmaglie could you please create a new 1.6.16 milestone for this?

PR #243, opened 2017-07-04
Add adjust_millis_forward() API to help with deep sleep processing

terrillmoore:
This patch fixes https://github.com/mcci-catena/ArduinoCore-samd/issues/1.  During a deep sleep, systick must be off. But some libraries use `millis()` to get an idea of real time. (In particular, this is true for https://github.com/mcci-catena/arduino-lmic -- otherwise, it loses track of the schedule.) Since `_ulTickCount` is local to delay.c, the appropriate correction is to add a new API `adjust_millis_forward()`. 

The submitted patch disables interrupts briefly. I have a different, tested, patch that instead causes the systick interrupt processing routine to do the adjustment. But I think the submitted approach is more clear, and is certainly more efficient. It would be great if this could be added to the mainline distribution. (Of course, this might also imply a similar change to the AVR core, and there are portability issues related to the function of the real-time clock which I have not addressed.)

With this patch, I have successfully entered deep sleep with an Adafruit Feather M0 for the roughly six-minute sleep between LoRaWAN transmissions in my application. Use of deep sleep doubles the battery life. 

Sample (somewhat clumsy use):
```c++
// Set an alarm in the future.
gRtc.SetAlarm(SLEEP_TIME_IN_SECONDS);

// Deep sleep until the alarm
gRtc.SleepForAlarm(gRtc.MATCH_HHMMSS, gRtc.SleepMode:IdleCpuAhbApb);

// At this point, approximately 1000 * SLEEP_TIME_IN_SECONDS ms have passed without interrupts. 
// Advance the value of millis() accordingly.
// (A more refined approach would take into account fractional seconds.)
adjust_millis_forward(SLEEP_TIME_IN_SECONDS * 1000);
```

### DIFF
--- a/cores/arduino/delay.c
+++ b/cores/arduino/delay.c
@@ -32,6 +32,29 @@ uint32_t millis( void )
   return _ulTickCount ;
 }
 
+// atomically add delta to _ulTickCount. Momentarily disables interrupts, assuming that
+// the systick interrupt is maskable.
+uint32_t adjust_millis_forward(uint32_t delta)
+{
+  // copy old interrupt-enable state to flags.
+  uint32_t const flags = __get_PRIMASK();
+
+  // disable interrupts
+  __set_PRIMASK(1);
+
+  // observe _ulTickCount, and advance it.
+  uint32_t const tickCount = _ulTickCount + delta;
+
+  // save _ulTickCount
+  _ulTickCount = tickCount;
+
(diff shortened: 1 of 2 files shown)


terrillmoore:
given no response, we just created our own BSP for our boards at https://github.com/mcci-catena/arduino-boards. Closing this.

PR #252, opened 2017-08-02
Use SAMD21 serial number for USB serial number prefix

sandeepmistry:
This is based on section 9.3.3 of the data sheet.

With this change, Blink is 1948 bytes larger on a MKR1000 (without change 9256 bytes, with change 11204 bytes).

Anyone know of a slimmer way to convert a register value to a hex string?


sandeepmistry:
I've made things slimmer using something based on `utoa` with leading zero padding and upper case conversion.

Blink on a MKR1000 is now `9760 bytes`, without the patch it's ` 9256 bytes` - so now a `504 byte` increase.

sandeepmistry:
Even simpler now with @cmaglie's suggestion, now Blink is down to `9336 bytes`, so just an 80 byte increase!

PR #253, opened 2017-08-23
Fix for pgm_read_ptr macro

nkrkv:
This is a fix for the same problem as for SAM platform. See https://github.com/arduino/ArduinoCore-sam/pull/43


cmaglie:
Thanks!

PR #254, opened 2017-08-23
Some updates on Stream and Print class.

cmaglie:


cores/arduino/Print.h
```diff
@@ -54,6 +57,10 @@ class Print
       return write((const uint8_t *)buffer, size);
     }
 
+    // default to zero, meaning "a single write may block"
+    // should be overriden by subclasses with buffering
+    virtual int availableForWrite() { return 0; }
```
sandeepmistry: Should we make `Uart:: availableForWrite()` virtual as well?

cmaglie: yes... good point...

sandeepmistry: Same for I2S, those are the only two spots in the repo.

sandeepmistry: Based on testing, adding the virtual doesn't seem to make a difference, we just needed to change the `size_t` to `int` in I2S.

PR #257, opened 2017-08-31
Add optional RTS + CTS support to UART

sandeepmistry:


cores/arduino/Uart.cpp
```diff
@@ -102,11 +129,30 @@ int Uart::peek()
...
+      digitalWrite(uc_pinRTS, LOW);
+    }
+  }
+
+  return c;
 }
 
 size_t Uart::write(const uint8_t data)
 {
+  if (uc_pinRTS != NO_RTS_PIN) {
```
cmaglie: why? should RTS influence only Rx?

cores/arduino/Uart.cpp
```diff
@@ -64,6 +84,13 @@ void Uart::IrqHandler()
 {
   if (sercom->availableDataUART()) {
     rxBuffer.store_char(sercom->readDataUART());
+
+    if (uc_pinRTS != NO_RTS_PIN) {
+      // if there is NOT enough space in the RX buffer, de-assert RTS
+      if (rxBuffer.availableForStore() < RTS_RX_THRESHOLD) {
+        digitalWrite(uc_pinRTS, HIGH);
```
cmaglie: `digitalWrite` may be slow, we should explore direct register access.

cores/arduino/Uart.cpp
```diff
@@ -102,11 +129,30 @@ int Uart::peek()
...
 }
 
 size_t Uart::write(const uint8_t data)
 {
+  if (uc_pinRTS != NO_RTS_PIN) {
+    // assert RTS
+    digitalWrite(uc_pinRTS, LOW);
+  }
+
+  if (uc_pinCTS != NO_CTS_PIN) {
```
cmaglie: mmmhhh there is no other way here? Ideally the CTS pin should mask the tx ISR, so you can always write in the buffer without blocking (until the buffer is full). What is not clear to me is if there is a way to trasparently re-enable the tx ISR.

cores/arduino/Uart.cpp
sandeepmistry: I've pushed a commit to leverage SERCOM h/w capabilities.

cores/arduino/Uart.cpp
sandeepmistry: Right, I misunderstood the RTS term - corrected!

cores/arduino/Uart.cpp
sandeepmistry: The latest commit addresses this.

sandeepmistry:
@cmaglie there's one open question, if we expose the `attachRts` and `attachCts` methods to the user, we need to validate the inputs are the right pins.

The other option is to pass then in the constructor.

sandeepmistry:
I've removed the `attachRts` and `attachCts` API's for now, and made RTS and CTS only configurable via the `Uart` constructor.

PR #260, opened 2017-09-07
Avoid stalling if uart buffer becomes full

facchinm:
The idea is borrowed from AVR core https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp#L236

This should allow other interrupts to fire while the buffer is getting (slowly) emptied.

### DIFF
--- a/cores/arduino/Uart.cpp
+++ b/cores/arduino/Uart.cpp
@@ -110,7 +110,11 @@ size_t Uart::write(const uint8_t data)
   if (sercom->isDataRegisterEmptyUART() && txBuffer.available() == 0) {
     sercom->writeDataUART(data);
   } else {
-    while(txBuffer.isFull()); // spin lock until a spot opens up in the buffer
+    while (txBuffer.isFull()) {
+      // Pretend an interrupt has happened and
+      // call the handler to free up space for us.
+      IrqHandler();
+    };
 
     txBuffer.store_char(data);
 


cmaglie:
Why it stalls? Are you trying to write from inside an ISR?

On the AVR side the IRQ handler is called only if the global interrupts are disabled, and only if the UART already emptied the internal shift register:

```C++
  // If the output buffer is full, there's nothing for it other than to 
  // wait for the interrupt handler to empty it a bit
  while (i == _tx_buffer_tail) {
    if (bit_is_clear(SREG, SREG_I)) {      <----- here
      // Interrupts are disabled, so we'll have to poll the data
      // register empty flag ourselves. If it is set, pretend an
      // interrupt has happened and call the handler to free up
      // space for us.
      if(bit_is_set(*_ucsra, UDRE0))    <------ and here
	_tx_udr_empty_irq();
    } else {
      // nop, the interrupt handler will free up space for us
    }
  }
```

I think that we should implement a similar check here

facchinm:
Yes, you caught me :smile: Indeed I was printing from an ISR; the additional check doesn't seem to fix the situation, nor forcefully rearming the `DataRegisterEmptyInterruptUART()` interrupt.

sandeepmistry:
Closing in favour of #262.

PR #267, opened 2017-10-15
 Adds additional analog inputs: A6-A11

Adminius:
A6 ~ D0
A7 ~ D1
A8 ~ D3
A9 ~ D4
A10 ~ D8
A11 ~ D9


mitchellpontague:
Of those pins only PA06 and PA07 are referenced to analog ground and intended to be used as analog inputs.
The attribute PIN_ATTR_ANALOG is only used to enable the DAC on analog writes. As is it will break PWM when referenced as A6-A11
The analogInputToDigitalPin macro is being used backwards, not mapping correctly, and mapping unintended pins.
The PIN_A6-PIN_A11 macros and A6-A11 constants should be added.

Adminius:
@mitchellpontague  why only PA06 and PA07? Source of this statement?
[analogInputToDigitalPin ](https://github.com/arduino/ArduinoCore-samd/pull/267/commits/72fbc56eb74b0b8d2f5263ec4d68d4eeedcfafa9)better?

mitchellpontague:
From the [datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/40001882A.pdf) on page 27 lists which pins are intended to be analog. Page 33 shows the voltage sources for each pin. However the only difference between the analog and io voltages is extra filtering.
analogInputToDigitalPin looks good.

Adminius:
grafik
@mitchellpontague PA08-PA11 can also be analog pins, but, as you say, without filtering.

Adminius:
@facchinm, @cmaglie hi, i know, it's a not a main function. But i would like to know if this PR in general will be accepted or if I have to make it in another way (like a external lib).

helmut64:
@Adminius important patch, I hope this gets merged soon.
I have also enabled the SWClock and SWData pins for the Arduino Zero which gives us two more pins when Debugging is not being used. As the two new Pins 44/45 belong to the same file, can you add them to your patch or should I do a separate patch:

Regards Helmut

The lines would be (at end, relaces 43, 44, 45 are new):
  // 43..45 - Alternate use of A0 (DAC output), 44 SDC, 45, SWD
  { PORTA,  2, PIO_ANALOG, PIN_ATTR_ANALOG, DAC_Channel0, NOT_ON_PWM, NOT_ON_TIMER, EXTERNAL_INT_2 }, // DAC/VOUT
  { PORTA, 30, PIO_PWM, PIN_ATTR_DIGITAL|PIO_SERCOM, No_ADC_Channel, NOT_ON_PWM, TCC1_CH0, EXTERNAL_INT_10 }, // SWC
  { PORTA, 31, PIO_PWM, PIN_ATTR_DIGITAL|PIO_SERCOM, No_ADC_Channel, NOT_ON_PWM, TCC1_CH1, EXTERNAL_INT_11 }, // SWD

Adminius:
i would say extra patch.
but there is/was a problem with SWCLK Pin:
[Using the SWD pins SWDIO and SWCLK for general in / out](http://www.avdweb.nl/arduino/samd21/samd21-variant.html)

helmut64:
I will do an extra patch, thank you for your advise.
My understanding is that the SWD/SWC pins are default connected to the debugger, switching them to GPIO is supported and works fine. I need extra simple GPIO pins that's why I do this. This is not a speciality on the D21 also STML4 can do this.

I still need to figure out who to check if a debugger is connect before using the pins. On STM there is a Debugger Interface active check available via (CoreDebug->DHCSR & 1)

On the SAMD/D21 I need need to figure out where the find the DHCSR address.

variants/arduino_zero/variant.h
```diff
@@ -94,6 +94,12 @@ extern "C"
 #define PIN_A3               (17ul)
 #define PIN_A4               (18ul)
 #define PIN_A5               (19ul)
+#define PIN_A6               (44ul)
+#define PIN_A7               (45ul)
+#define PIN_A8               (46ul)
+#define PIN_A9               (47ul)
+#define PIN_A10              (48ul)
+#define PIN_A11              (49ul)
 #define PIN_DAC0             (14ul)
```
sandeepmistry: These additional new analog pins will break the logic in: https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/wiring_analog.c#L132-L134

Adminius: this logic is for using analogRead function with 0-5 instead of A0-A5...
so that means:
```
 if (pin < A0) {
    pin += A0;
  }else if(pin > 5 && pin <= 11){
    pin += 38;
 }
```
@sandeepmistry is it ok for you?

Adminius:
this PR is outdated. new: #374

PR #273, opened 2017-11-09
No fixed value for USB power current

awatterott:
The USB power config is currently a fixed value of 500mA. On most devices this has no effect, but on iPads all devices with a power value/consumption over 20mA will not work. So it would be good to have a way to change this value without modifying files from the Arduino Core.


nevvkid:
+1 For this change. It would help me with my current project.

facchinm:
@ArduinoBot build this please

awatterott:
Hi Martino,
I think it can be added also to the other cores. Should I make a PR there, too?

facchinm:
@awatterott sure! I'm merging this one in the meantime. Thanks!

PR #275, opened 2017-11-10
Fixed the iManufacturer and iProduct HID string descriptor functions

adammunich:
iManucaturer and iProduct were not properly written for HID devices, as the setup.wlength struct element did not represent the real length of the string descriptor.

This caused HID driver errors on Windows AMD_64.

String descriptors are supposed to be in unicode, so 2*sizeof(string) gives the proper length.


facchinm:
LGTM, testing it!

PR #276, opened 2017-11-16
Added c++ linker command to allow to include libstdc++ when linking.

helmut64:
This is needed otherwise the template C++ container support
are missing. The "compiler.c.elf.cmd" specifies the linker,
the C linker was wrong here and does not work for true C++ apps.

This change does not hurt because when no libstdc++ features are
being used there is no change.

Changed the Arduino function map() into Arduino_map(). This is needed
otherwise it clashes with the C++ template class. e.g.: #include <map>

PS: I also added the same changes into the 32-bit Arduino ESP32 release.

### DIFF
--- a/cores/arduino/WMath.cpp
+++ b/cores/arduino/WMath.cpp
@@ -52,7 +52,7 @@ extern long random( long howsmall, long howbig )
   return random(diff) + howsmall;
 }
 
-extern long map(long x, long in_min, long in_max, long out_min, long out_max)
+extern long Arduino_map(long x, long in_min, long in_max, long out_min, long out_max)
 {
   return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
 }
--- a/cores/arduino/WMath.cpp
+++ b/cores/arduino/WMath.cpp
@@ -52,7 +52,7 @@ extern long random( long howsmall, long howbig )
   return random(diff) + howsmall;
 }
 
-extern long Arduino_map(long x, long in_min, long in_max, long out_min, long out_max)
+extern long map(long x, long in_min, long in_max, long out_min, long out_max)
 {
   return (x - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
 }
--- a/cores/arduino/WMath.h
+++ b/cores/arduino/WMath.h
@@ -22,7 +22,7 @@
 extern long random( long ) ;
 extern long random( long, long ) ;
 extern void randomSeed( uint32_t dwSeed ) ;
-extern long map( long, long, long, long, long ) ;
+extern long Arduino_map( long, long, long, long, long ) ;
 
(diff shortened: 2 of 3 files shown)


helmut64:
The other map is the regular C++ libstdc++ map which is being used in all C++ books, C++ projects, etc.. C++ is now here with this for more than 20 years. It is a fault by Arduino to use this name. It is also little being used compared to the C++ map.

helmut64:
PS: I also checked the memory size of apps for the D21 SAMD platform, there is absolutely no difference with this patch.

helmut64:
I believe now it should work.

helmut64:
Why does it take several months to review and integrate my enhancement.

PaulStoffregen:
> Changed the Arduino function map() into Arduino_map().

How does this work?  Looking at the files changed tab, I see a define, but nothing else which would allow the huge number of programs people have written with map() to compile and actually work.  Did I miss something?

Hopefully you can understand how a change that breaks (or likely breaks) one of Arduino's very widely used features would never be accepted?

helmut64:
Hi Paul, the map problem is resolved by using the complete C++ namespace name std::map for C++. The remaining issue is the wrong linker statement. It can be fixed via

<< compiler.c.elf.cmd=arm-none-eabi-gcc
>> compiler.c.elf.cmd=arm-none-eabi-g++

With the ESP32 this is no problem,  for the the D21/SAMD platform switching to the g++ for linking will resolve the STL C++ libraries when used.

Do you have merge rights?

It is just disappointing that I do a fix, for the ESP32 it took a few days to get it merged/resolved, for the SAMD nothing happens for 7 month.

Regards Helmut

cmaglie:
I tried this with the example:

```
void setup() {
  int x = map(1,2,3,4,5);
}
void loop() { }
```

and it fails with:

```
/tmp/arduino_modified_sketch_608153/sketch_jun27a.ino: In function 'void setup()':
sketch_jun27a:3:24: error: 'Arduino_map' was not declared in this scope
   int x = map(1,2,3,4,5);
                        ^
/tmp/arduino_modified_sketch_608153/sketch_jun27a.ino:3:7: warning: unused variable 'x' [-Wunused-variable]
   int x = map(1,2,3,4,5);
       ^
exit status 1
'Arduino_map' was not declared in this scope
```

are you sure that the PR is complete?
I see that there are a bunch of tentative commits that are undone later, maybe you want to clean up the history?

cmaglie:
Oh I see, I guess this is a leftover:

https://github.com/arduino/ArduinoCore-samd/pull/276/files#diff-9ad272c11f2032a6e0231ebda4438a04R31

so just changing `gcc` with `g++` for linking should be enough?

helmut64:
Yes, just changing the linker to use g++
Regards Helmut

PR #280, opened 2017-11-20
samd21e sercom compatibility

mitchellpontague:
Make sercom4 and sercom5 optional to allow compilation of samd21e variants.


facchinm:
LGTM. @mitchellpontague could you get rid of the first two diff lines (about I2C) so we can merge it as is? Thanks

sandeepmistry:
@mitchellpontague thanks!

PR #281, opened 2017-12-08
Avoid read-modify-write when enabling SERCOM DRE interrupt

SodaqMoja:
The disable interrupt already does it like this. Just write the value of the DRE bit, no read-modify-write operation is needed.


sandeepmistry:
@SodaqMoja thanks!

PR #282, opened 2017-12-11
Allow General Calls on I2C bus (aka broadcast)

agdl:
This allow to call a Wire.begin(address, true) to enable general call (broadcast messages on address 0x00) feature on SAMD21 based devices. All the legacy calls to Wire.begin(address) are ok since the extra parameter is false by default

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -400,7 +400,7 @@ void SERCOM::disableWIRE()
   }
 }
 
-void SERCOM::initSlaveWIRE( uint8_t ucAddress, bool broadcast )
+void SERCOM::initSlaveWIRE( uint8_t ucAddress, bool enableGeneralCall )
 {
   // Initialize the peripheral clock and interruption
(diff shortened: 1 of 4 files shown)


cores/arduino/SERCOM.cpp
```diff
@@ -400,17 +400,20 @@ void SERCOM::disableWIRE()
   }
 }
 
-void SERCOM::initSlaveWIRE( uint8_t ucAddress )
+void SERCOM::initSlaveWIRE( uint8_t ucAddress, bool broadcast )
```
facchinm: Should we call the parameter `listen_on_broadcast` or something similar? @sandeepmistry

sandeepmistry: Maybe `enableGeneralCall`?

agdl:
@lupalby I think you will like this too ;) https://github.com/agdl/WireUpdate

lupalby:
@agdl That's a cool feature! I will look into it

Issue #259, opened 2017-09-04
Incorrect I2C SCL frequency


MartinL1:
I observed that the I2C SCL clock in fast mode (400kHz) is only running at 357kHz, despite using 2k2 pull-up resistors that provide a relatively fast rise time.

After some investigation it turns out that the SERCOM::InitMasterWire() function (in the file SERCOM.cpp) is using the incorrect formula for the BAUD register.

The current fomula appears to be taken from Section 25 SERCOM - Serial Communication Interface in the SAMD21 datasheet:

BAUD = SystemCoreClock (48MHz) / (2 * baud rate) - 1

...however, the correct formula should be the one from Section 28 SERCOM - Inter-Integrated Circuit:

BAUD = SystemCoreClock (48MHz) / (2 * SCL_Frequency) - 5 - (SystemCoreClock (48MHz) * T(rise) / 2)

...where T(rise) is the time taken for the pull-up resistor to pull the SCL line high, (and will very much depend on the resistor values used).

Using 2k2 pull-up resistors I measured the rise time as 90ns, therefore the BAUD value in this instance should be:

BAUD = 48MHz / (2 * 400000) - 5 - (48MHz * 90ns / 2) = 52.84

Using a BAUD value around 52 or 53 gives a SCL frequency close to 400kHz.

Adminius:
do you have also checked if other frequencys are ok? (e.g. 1MHz? )

MartinL1:
The SAMD21 datasheet states that for standard I2C fast mode plus at 1MHz, "requires a nominal high to low SCL ratio 1:2". This involves setting the BAUDLOW bitfield in the BAUD register. 

In this case the SCL frequency is determined by the formula:

SCL_Frequency = SystemCoreClock (48MHz) / (10 + BAUDLOW + BAUD + (SystemCoreClock (48Mhz) * T(rise)))

As it currently stands the Wire library doesn't take this into account.

deladriere:
@Adminius 
I measure  :  
350 kHz when I set the clock to 400 kHz
735 kHz when I set the clock to 1 MHz
1.16 MHz when I set the clock to 2 MHz
1.45 MHz when I set the clock to 3 MHz
using Wire.setClock(freq);

sandeepmistry:
Fixed via #287.

patryk02:
I have a similar problem with my MKRzero board. 
[...]

PR #287, opened 2017-12-22
Correct I2C frequency calculation formula

sandeepmistry:
As discussed in #259.

Also alllow variant to override the default rise time in nanoseconds via define.

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -19,6 +19,12 @@
 #include "SERCOM.h"
 #include "variant.h"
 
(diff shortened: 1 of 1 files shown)


sandeepmistry:
Hi @MartinL1, @Adminius, @deladriere,

Could you please try out these changes and see if it addresses the discussion in issue #259. The rise time is based on the 4.7K pull up resistors all the Arduino.cc SAMD boards. Other variants can override the default value if needed.

sandeepmistry:
These changes were tested with a MKR Zero and MKR WAN1300, using a OLED display and ECC508 over the I2C bus. A Saleae logic analyzer was used to observe traffic on the I2C bus.

PR #289, opened 2017-12-22
Overload beginTransmission and requestFrom to support using an external buffer

sandeepmistry:
Replacement for #269, with updated API's based on the discussion with @lupalby. `Wire. requestFrom(...)` and `Wire.beginTransmission(...)` are overloaded to support external buffers.

The external buffer for TX can optionally be partially or fully pre-populated with data. The existing `Stream` and `Print` API's can continue to be used to populate the external buffer.


PaulStoffregen:
Some time ago we all discussed an "addMemory" API to be added to HardwareSerial, Wire, and other libs.

Is that idea gone now?

sandeepmistry:
Closing in favour of #298 for now.

After internal discussions we have decided not to add additional API's to the `Wire` class. Instead PR #298 increases the Wire buffer size of 256 for both TX and RX.

Issue #293, opened 2018-01-15
SerialUSB cause memory leak after LowPower.sleep(), using ArduinoLowPower lib. samd21 cpu


Terryp44:
Every wakeup causes some memory loss, eventually causing CPU reset. However if USB is not plugged in, no memory is lost.
using USBDevice.detach(); and USBDevice.attach();, open/close port does not help.
Using samd version 1.6.17
Test RAM memory usage via:
extern "C" char *sbrk(int i);
int FreeRam (void)
{
  char stack_dummy = 0;
  return &stack_dummy - sbrk(0);
}

Terryp44:
After reconnecting the SerialUSB port to the PC. samd CPU ram memory decreases by 592 bytes every sleep cycle.

Terryp44:
After reconnecting the SerialUSB port to the PC. samd CPU ram memory decreases by 592 bytes every sleep cycle.

facchinm:
Hi @Terryp44 ,
I pushed #295 to solve the issue. Can you test the produced json and confirm that the standby "dance" doesn't leak anymore? Thanks!

PR #295, opened 2018-01-17
Don't reallocate USB buffers if already allocated

facchinm:
USB Configuration was meant to run only once, but if the board comes back from standby the host can reconfigure the device again.
Probably a cleaner patch could be free()-ing the buffers on standby() to release the memory but at least we don't leak anymore.

Fixes https://github.com/arduino/ArduinoCore-samd/issues/293

### DIFF
--- a/cores/arduino/USB/USBCore.cpp
+++ b/cores/arduino/USB/USBCore.cpp
@@ -454,7 +454,9 @@ void USBDeviceClass::initEP(uint32_t ep, uint32_t config)
 	}
 	else if (config == (USB_ENDPOINT_TYPE_BULK | USB_ENDPOINT_OUT(0)))
 	{
-		epHandlers[ep] = new DoubleBufferedEPOutHandler(usbd, ep, 256);
+		if (epHandlers[ep] == NULL) {
+			epHandlers[ep] = new DoubleBufferedEPOutHandler(usbd, ep, 256);
+		}
 	}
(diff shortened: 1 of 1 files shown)


Terryp44:
Hi, Thanks a lot for making this change. The SerialUSB port will only work after the first sleep cycle. The next sleep cycle, the SerialUSB port does not communicate.  Reconnecting the USB cable does not fix it either. A board soft reset allows port to work again. however Putty does not lose connection ....just does not communicate. Then if the USB cable is unplugged and then plugged back in, putty displays a writing error.but device manager shows Arduino zero is connected.

facchinm:
I'm using this sketch to test the fix
```arduino
#include "ArduinoLowPower.h"

extern "C" char *sbrk(int i);
int FreeRam (void)
{
char stack_dummy = 0;
return &stack_dummy - sbrk(0);
}

void setup() {
  SerialUSB.begin(115200);
  while (!SerialUSB){}
}

void loop() {
  // put your main code here, to run repeatedly:
  SerialUSB.println(FreeRam());
  LowPower.sleep(1000);
}
```
and I can get the prints before and after sleep(). My host PC is Linux so the different behaviour can be easily explained with the different USB stack.
Are you using version 1.2.0 of LowPower library?

Terryp44:
Thanks Again, Well same behavior as before, using your sketch. I am using custom hardware, so there could be something there. Our LowPower Lib is the same. Or my PC just does not like re-connecting after CPU sleep.
Thanks for your help. the memory leak is fixed!!!

PR #306, opened 2018-02-17
Fix CDC_ENABLED

gdsports:
Undefining CDC_ENABLED triggered a compiler error. With this fix, the CDC ACM configuration can be disabled.


sandeepmistry:
@gdsports thank you!

PR #314, opened 2018-03-10
Larger Buffer RingBuffer.h

d3str0y3rport:
The increase of the buffer size to handle larger streams like the ones from GPS.

### DIFF
--- a/cores/arduino/RingBuffer.h
+++ b/cores/arduino/RingBuffer.h
@@ -27,8 +27,9 @@
 // using a ring buffer (I think), in which head is the index of the location
 // to which to write the next incoming character and tail is the index of the
 // location from which to read.
-//#define SERIAL_BUFFER_SIZE 64
+#ifndef defined(SERIAL_BUFFER_SIZE)//allows to define the user the buffer size
 #define SERIAL_BUFFER_SIZE 128//Double the buffer size to get larger streams like from GPS
+#endif
 
 template <int N>
 class RingBufferN
--- a/cores/arduino/RingBuffer.h
+++ b/cores/arduino/RingBuffer.h
@@ -27,7 +27,8 @@
 // using a ring buffer (I think), in which head is the index of the location
 // to which to write the next incoming character and tail is the index of the
 // location from which to read.
-#define SERIAL_BUFFER_SIZE 64
+//#define SERIAL_BUFFER_SIZE 64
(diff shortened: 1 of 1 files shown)


Adminius:
why not let user define his own buffer size?

> #ifndef defined(SERIAL_BUFFER_SIZE)
> #define SERIAL_BUFFER_SIZE 64
> #endif

tuxedo0801:
@cmaglie @sandeepmistry Any comments on this?

Adminius:
@d3str0y3rport sorry, i make a mistake:
1. possibility:
`#ifndef SERIAL_BUFFER_SIZE`
2. possibility
`#if !defined(SERIAL_BUFFER_SIZE)`
see also: https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/HardwareSerial.h#L42

cores/arduino/RingBuffer.h
```diff
@@ -27,7 +27,9 @@
 // using a ring buffer (I think), in which head is the index of the location
 // to which to write the next incoming character and tail is the index of the
 // location from which to read.
-#define SERIAL_BUFFER_SIZE 64
+#ifndef defined(SERIAL_BUFFER_SIZE)//allows to define the user the buffer size
+#define SERIAL_BUFFER_SIZE 128//Double the buffer size to get larger streams like from GPS
```
sandeepmistry: I'd be up for bumping this to 256 by default.

The comment about doubling the buffer size should be removed before this pull request can be merged.

keestux:
Can you also show how and where you would define a different value? You'd have to define SERIAL_BUFFER_SIZE in the right place.

And what about if you have an application with multiple modules (one .ino and one or more .cpp)? Each module may or may not include Arduino.h (or Uart.h, or RingBuffer.h). If you don't set the same SERIAL_BUFFER_SIZE everywhere then you're screwed.

keestux:
Another note. Each Uart has 2 ringbuffers. That means the data size increases quite a bit if the variant has 3 or 4 Serial devices.

sandeepmistry:
Closing in favour of: 
https://github.com/arduino/ArduinoCore-samd/commit/da9bfe8cf9aa92dc0e4f480be44fcecd5efe1604

sandeepmistry:
@keestux 

> Can you also show how and where you would define a different value?

This would have to be changed in the boards.txt entry of the variant you are using.

PR #315, opened 2018-03-15
[SPI] Fix interrupt mask to block

ggajoch:
Digital pin number is not equivalent to SAMD interrupt number (in EIC module).
Mapping is provided by `g_APinDescription` pin map. 

However, this solution assumes that interrupt number is equal to digital pin number (which holds true for SAMD devices):
https://github.com/arduino/ArduinoCore-samd/blob/d2b2ff928ebf9b34a1621f50949e1f5cf088b21d/cores/arduino/Arduino.h#L124

Fix for https://github.com/arduino/ArduinoCore-samd/issues/316.


sandeepmistry:
@ggajoch thank you!

Issue #310, opened 2018-03-02
I2S sampling freq not correct for 8-bit @ 8 kHz


andrewjfox:
Hi,

I've been trying to get audio from a modem using the I2S/PCM interface and having some troubles getting accurate clock frequencies.

I'm trying to read single 16bit words from modem @ 8kHz, by configuring the I2S as 8bit slots, since I see in the library it's expecting stereo so has min 2 slots. Following is code to init device, note that I am adjusting clock polarity and FS width to suit modem. 
```
 I2S.onReceive(ISR_onI2SReceive); 
 
  if(!I2S.begin(I2S_PHILIPS_MODE, 8000, 8)) return false;  
  
  /*  SAMD21 I2S peripheral samples data on +ve clock edge, however modem data is valid on -ve edge
   *  if SAMD is master then can invert output clock, this only changes output, so internal operation is unchanged
   *  modem requires FS to be 1 bit width (in slave mode)
  */
  i2s_disable();
  REG_I2S_CLKCTRL0 &= ~I2S_CLKCTRL_FSWIDTH_Msk;  //clear FS width
  REG_I2S_CLKCTRL0 |= I2S_CLKCTRL_SCKOUTINV | (I2S_CLKCTRL_FSWIDTH_BIT_Val << I2S_CLKCTRL_FSWIDTH_Pos);  //invert clock output, FS 1 bit wide (TDM mode)
  i2s_enable();

```
For this setup, with 8kHz sample rate I would expect SCLK to be 8000 * 2 * 8 = 128kHz, however when I monitor the I2S lines using logic analyser, the clock is around 400kHz, and FS is around 25kHz. See attached screen capture,
![logic_i2s_8khz_8bit](https://user-images.githubusercontent.com/14087194/36883769-b3b7c88e-1e30-11e8-81b1-82d183ef4a06.jpg)
 
If I have it configured for 8kHz and 16bit data, then clock and frame are at 256kHz and 8kHz resepctively, as expected.

I had a look at the driver code but the clock setting is a bit complex at first glance so wasn't really able to locate where the issue is.

Regards
Andrew

andrewjfox:
Hi @sandeepmistry , just bumping this to see if I can get a response, and I think you've been involved in the I2S stuff before.

andrewjfox:
Please! Is anyone able to validate that this is a bug in the I2S library?

sandeepmistry:
[...]

PR #317, opened 2018-03-22
I2S: Use 8 MHz oscillator source if 48MHz divider does not fit in 8 bits

sandeepmistry:
Fixes sketch from @andrewjfox in https://github.com/arduino/ArduinoCore-samd/issues/310#issuecomment-375144902

The divider can have a max resolution of 8-bits since we are using generator 3:

screen shot 2018-03-22 at 3 11 50 pm

These changes use the 8 MHz oscillator as the clock source instead of the 48 MHz DFLL if the divider can't fit in 8-bits.

### DIFF
--- a/libraries/I2S/src/I2S.cpp
+++ b/libraries/I2S/src/I2S.cpp
@@ -398,15 +398,24 @@ void I2SClass::onReceive(void(*function)(void))
 
 void I2SClass::enableClock(int divider)
 {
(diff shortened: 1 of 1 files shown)


andrewjfox:
Just tested and getting correct clock freq for 8 bits @  8kHz sampling

sandeepmistry:
@andrewjfox great, thanks for trying and your feedback!

PR #329, opened 2018-06-29
[UART] add sendBreak(duration) API

sandeepmistry:
Note: `SERCOM_USART_CTRLA_FORM` needs to be set to auto baud to enable break detection.

Since there is no easy way on the SAMD to detect the break duration, the break value is set to 0xffff when a break is detected, otherwise -1 when no break is detected.


sandeepmistry:
Closing this for now, we can do something outside the core for now.

PR #330, opened 2018-07-04
Added defines to identify MKR pin layout

cmaglie:
/cc @PaulStoffregen @agdl @facchinm


cmaglie:
> It does only work if the library includes "Arduino.h" directly or indirectly, thus it's not a real replacement for ARDUINO_SAMD_?

Right, now I've put it into boards.txt so it should be always available. BTW it is not a replacement for `ARDUINO_SAMD_xxxxx`, in the sense that this tells you about the board format not the board model.

PR #331, opened 2018-07-13
Added alternate ports 44,45 to make the SWCLK and SWDIO pins available

helmut64:
Added alternate ports 44,45 to make the SWCLK and SWDIO pins available 
when no debugger is being used.
BTW. The presence of a debugger can be checked via:
if (DSU->STATUSB.bit.DBGPRES) ...
@cmaglie it would be great if this patch can be reviewed and merged. Thank you for your help.

### DIFF
--- a/variants/arduino_zero/variant.cpp
+++ b/variants/arduino_zero/variant.cpp
@@ -96,6 +96,10 @@
  * |            | GND              |        |                 |
  * | 42         | AREF             |  PA03  |                 | EIC/EXTINT[3] *[ADC|DAC]/VREFA ADC/AIN[1] PTC/Y[1]
  * +------------+------------------+--------+-----------------+--------------------------------------------------------------------------------------------------------
+ * | 43         |                  |  PA02  |                 | Alternate use of A0 (DAC output)
+ * | 44         |                  |  PA30  |                 | SWCLK, alternate use EXTINT[10] TCC1/WO[0] SERCOM1/PAD[2]
+ * | 45         |                  |  PA31  |                 | SWDIO, alternate use EXTINT[11] TCC1/WO[1] SERCOM1/PAD[3]
+ * +------------+------------------+--------+-----------------+--------------------------------------------------------------------------------------------------------
  * |            |32.768KHz Crystal |        |                 |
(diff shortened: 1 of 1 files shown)


helmut64:
@cmaglie @facchinm  I hope my patch can be reviewed soon.
Also the patch #267 from @Adminius is as well of interest and hanging around for 10 month or so.

I love to fix issues and bring enhancements, however my motivation goes to zero when it takes 9 month to review 5 lines. Lets collaborate and bring this forward.

Warm regards from the sunny  Hannover, Germany.
Helmut

sandeepmistry:
Hi @helmut64,

First, thank you for submitting this pull request!

Is the intent to use the tiny header on the Arduino Zero board or is this for another board? Are you running out of pins that are broken out on the standard Arduino form factor headers?

helmut64:
Hi Sandeep, this patch is indeed when you run out of pins on the Arduino Zero boards, certainly you need to attach cables to the pins.

We developed also a own Arduino Zero board with LoRa and Debugger and Display Interface where we need this additional pins. Of course for our solution we check if the SWD pin is in use with a debugger and don't use it when busy. Here is our board: http://www.radioshuttle.de/longra/board/
It is a great benefit for our users that they can just specify Arduino Zero in the board manager.
Regards Helmut

sandeepmistry:
> We developed also a own Arduino Zero board with LoRa and Debugger and Display Interface where we need this additional pins.

So for this scenario I would recommend you create your own core/variant, which depends on the official SAMD core: https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5-3rd-party-Hardware-specification#referencing-another-core-variant-or-tool

Adminius:
bords.txt will be overwritten after each update... and "own" core should be maintained/updated after arduino-core update... 
it is a good solution for testing, but terrible for normal operation :/

helmut64:
We designed the D21 Board to be compatible with the zero board, the patch applies to the regular Arduino Zero as well to our D21 board. We have done the Zero compatibility by purpose.

I feel that the patch enabling the pins should be ok.
PS: For different MCUs and features I agree  that a custom variant is ok, however we are not in the volume business that it makes much sense.

helmut64:
Danke Cristian.

PR #338, opened 2018-07-30
CPX fixes: add &MI_00 to all devices so CPLAYBOOT is not hidden on Windows 10

### LINKED ISSUE #336: Change in Windows Drivers breaking Circuit Playground Express Bootloader
Hey @cmaglie 

Several people have recently had trouble using CPX on Windows 10: CPLAYBOOT disk drive doesn't show up. This seems to happen after they install Arduino.
Example: https://forums.adafruit.com/viewtopic.php?f=58&t=138649

We know from experience that you shouldn't install the .inf file on windows 10, and let windows 10 use the default CDC driver. 

What's odd is that this wasn't an issue till recently, did the install procedure change?

dhalbert:
File naming has also been normalized to match the other drivers in this directory: all lower case, and the `.inf` and `.cat` files now have the same name.

Fixes #336.

@ladyada and I have tested this on Windows 7 and Windows 10.

Could you make a new minor release after accepting this PR? Thanks!

(EDIT: also version number of driver has been bumped and made more rational)


dhalbert:
@cmaglie We just found out that webUSB beta is available (again) for MakeCode, and this driver will interfere with that. We'll have another PR that fixes that soon, if the 1.6.20 release is not imminent. We need to suppress Circuit Playground Express driver installation completely on Windows 10, probably in the `post_install.bat` similar to what you suggested in #336.

The changes in this PR are still fine: the .inf is an improvement in general on the previous one.

tagging @ladyada

PR #341, opened 2018-08-03
[UART] Add frame and parity error handling

Rocketct:
Introduced Frame Error check for the received byte

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -137,7 +137,7 @@ bool SERCOM::isUARTError()
 
 bool SERCOM:: isParityOrFrameError()
 {
-  return sercom->USART.STATUS.bit.FERR | sercom->USART.STATUS.bit.PERR ;
+  return (sercom->USART.STATUS.reg & (SERCOM_USART_STATUS_FERR |SERCOM_USART_STATUS_PERR));
 }
 
 void SERCOM::acknowledgeUARTError()
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -159,13 +159,13 @@ bool SERCOM::isFrameErrorUART()
 
 void SERCOM::clearFrameErrorUART()
 {
-  //clear FERR bit writing 1  status bit
+  // clear FERR bit writing 1 status bit
   sercom->USART.STATUS.bit.FERR = 1;
 }
 
 void SERCOM::clearParityErrorUART()
 {
-  //clear FERR bit writing 1  status bit
+  // clear FERR bit writing 1 status bit
   sercom->USART.STATUS.bit.PERR = 1;
 }
 
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -137,7 +137,7 @@ bool SERCOM::isUARTError()
 
 bool SERCOM:: isParityOrFrameError()
 {
-  return (sercom->USART.STATUS.reg & (SERCOM_USART_STATUS_FERR |SERCOM_USART_STATUS_PERR));
+  return (sercom->USART.STATUS.reg & (SERCOM_USART_STATUS_FERR | SERCOM_USART_STATUS_PERR));
 }
 
 void SERCOM::acknowledgeUARTError()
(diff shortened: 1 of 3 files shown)


cores/arduino/Uart.cpp
```diff
@@ -91,8 +91,11 @@ void Uart::flush()
 void Uart::IrqHandler()
 {
   if (sercom->availableDataUART()) {
-    rxBuffer.store_char(sercom->readDataUART());
-
+    sercom->isFrameErrorUART();
```
sandeepmistry: This line doesn't do anything useful, just read the error and ignore.

Should we also clear the frame error?

cores/arduino/Uart.cpp
```diff
@@ -91,8 +91,11 @@ void Uart::flush()
...
+    uint8_t c=sercom->readDataUART();
+    if( !sercom->isUARTError()){
```
sandeepmistry: I think this should also check the error type here, from the data sheet not all errors will have data associated with them.

cores/arduino/Uart.cpp
```diff
@@ -91,10 +91,13 @@ void Uart::flush()
 void Uart::IrqHandler()
 {
   if (sercom->availableDataUART()) {
-    sercom->isFrameErrorUART();
-    uint8_t c=sercom->readDataUART();
-    if( !sercom->isUARTError()){
+    uint8_t c = sercom->readDataUART();
+    if (!sercom->isUARTError()) {
```
sandeepmistry: I still think this condition is incorrect, please look at the data sheet again, not all error situations should result in us dropping the byte.

cores/arduino/SERCOM.cpp
```diff
@@ -135,6 +135,11 @@ bool SERCOM::isUARTError()
   return sercom->USART.INTFLAG.bit.ERROR;
 }
 
+bool SERCOM:: isParityOrFrameError()
+{
+  return sercom->USART.STATUS.bit.FERR | sercom->USART.STATUS.bit.PERR ;
```
sandeepmistry: here i think we can optimize this by doing a single read of `.reg` instead of `.bit` and comparing the bit mask

cores/arduino/SERCOM.h
```diff
@@ -164,6 +164,8 @@ class SERCOM
 		int writeDataUART(uint8_t data) ;
 		bool isUARTError() ;
 		void clearFrameErrorUART() ;
+		void clearParityErrorUART() ;
+		bool  isParityOrFrameError();
```
sandeepmistry: nit pick: extra space here

cores/arduino/Uart.cpp
```diff
@@ -91,13 +91,17 @@ void Uart::flush()
...
 {
   if (sercom->availableDataUART()) {
-    uint8_t c = sercom->readDataUART();
-    if (!sercom->isUARTError()) {
-        rxBuffer.store_char(c);
+
+    if (!sercom->isParityOrFrameError()) {
+        rxBuffer.store_char(sercom->readDataUART());
     } else {
+      sercom->readDataUART();
```
sandeepmistry: I think we should add a comment above this live, saying read and discard the data, as it's invalid

cores/arduino/Uart.cpp
```diff
@@ -91,13 +91,17 @@ void Uart::flush()
...
       if (sercom->isFrameErrorUART()) {
         sercom->clearFrameErrorUART();
       }
+      if (sercom->isFrameErrorUART()) {
+              sercom->clearFrameErrorUART();
```
sandeepmistry: there's some weird spacing here? tabs vs spaces?

sandeepmistry:
@cmaglie can we please try to get this in before the next SAMD core release. Let me know if you have any comments.

cores/arduino/Uart.cpp
```diff
@@ -91,7 +91,23 @@ void Uart::flush()
...
+    } else {
+      // read the invalid data and discard it
+      sercom->readDataUART();
+      
+      // clear any errors that are set
+      if (sercom->isFrameErrorUART()) {
+        sercom->clearFrameErrorUART();
+      }
+      
+      if (sercom->isFrameErrorUART()) {
```
sandeepmistry: I think this block should check for parity error, not frame error again.

sandeepmistry:
Closing in favour of #349.

Issue #342, opened 2018-08-04
Native USB Host device failures


gdsports:
There are four USB devices that fail on the Zero native USB port but work on Due. The MouseController and KeyboardController examples are used for testing.

```
                                Due     Zero    Zero-patch
                                ===     ====    ==========
Logitech Marble mouse           Pass    Fail    Pass
Generic no-name mouse           Pass    Pass    Pass

```

The non-keyboard devices emulate USB keyboards.

```
                                Due     Zero    Zero-patch
                                ===     ====    ==========
Generic Barcode reader          Pass    Fail    Pass
Generic RFID reader small       Pass    Fail    Pass
Generic RFID reader large       Pass    Fail    Pass
Generic Numeric keypad          Pass    Pass    Pass
Generic 101 Keyboard            Pass    Pass    Pass
```

sandeepmistry:
Closed via #343.

PR #343, opened 2018-08-04
Fix USB Host failures

gdsports:
Fix for Issue #342.

Due does not do a Bus Reset before setting the bus address so this matches Due. 

One device sends a HID report descriptor that is more than 64 bytes so increasing the buffer to 128 should cover more devices.

### DIFF
--- a/libraries/USBHost/src/hidboot.h
+++ b/libraries/USBHost/src/hidboot.h
@@ -311,10 +311,6 @@ uint32_t HIDBoot<BOOT_PROTOCOL>::Init(uint32_t parent, uint32_t port, uint32_t l
     }
 
 
-	// Reset
-	//UHD_BusReset();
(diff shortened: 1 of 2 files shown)


libraries/USBHost/src/hidboot.h
```diff
@@ -312,8 +312,8 @@ uint32_t HIDBoot<BOOT_PROTOCOL>::Init(uint32_t parent, uint32_t port, uint32_t l
 
 
 	// Reset
-	UHD_BusReset();
-	while( Is_uhd_starting_reset() ) {}
+	//UHD_BusReset();
+	//while( Is_uhd_starting_reset() ) {}
```
sandeepmistry: @gdsports @facchinm is we don't need these lines any more, let's remove instead of commenting out

Issue #271, opened 2017-10-20
memory overwrite in samd21_host.c


alanambrose:
I believe this fix is needed in UHD_Init.  I tracked down an overwrite in some memory I had allocated which was being stamped on by the loop below with the uint32 cast (because sizeof() returns # bytes and the uint32 casting counts up in 32-bit **words**). The replacement memset works fine.:

	// ACA - memory overwrite!!
	memset(usb_pipe_table,0,sizeof(usb_pipe_table));
	/*
	for (i = 0; i < sizeof(usb_pipe_table); i++)
	{
		(*(uint32_t *)(&usb_pipe_table[0] + i)) = 0;
	}
	*/

sandeepmistry:
Hi @alanambrose,

The suggestion looks good to me, would you like to submit a pull request for it?

sandeepmistry:
Closed via https://github.com/arduino/ArduinoCore-samd/pull/355.

PR #355, opened 2018-09-02
Fix memory overwrite in USBHost

gdsports:
Fix issue #271.


cmaglie:
Manually merged with a small indent fix: 9ac80d9

PR #357, opened 2018-09-06
Added support for Minitronics v2.0

brupje:
Added support for Minitronics v2.0, based on the SAMD21J18


facchinm:
Hi @brupje ,
in this repo we only maintain official Arduino boards; since you don't have any difference to be applied for `arduino` core but only `variant` additions I suggest you to produce a derived core and make it available through https://github.com/arduino/Arduino/wiki/Unofficial-list-of-3rd-party-boards-support-urls .
I can give you advice on this if you encounter any problem; in the meantime I'm closing the PR

PR #369, opened 2018-10-31
Copy of samd21 usb.h for use in samdr21

drogge:
Femtoduino uses the samd21 version of SAMD21_USBDevice.h but ends up including the samr21 version of usb.h. This change places a copy of the samd21 usb.h in the bootloaders samr21 include direrectory. Fixes issue # 26.


facchinm:
Wrong target remote I think :slightly_smiling_face:

drogge:
Probably but I don't know if there's a sam R21 specific arduino core.  I'm going out on a limb here but I think the difference between the D21 and R21 may be that the R21 has bluetooth.

Issue #279, opened 2017-11-19
Adding a NUL byte corrupts String administration


SodaqMoja:
https://github.com/arduino/ArduinoCore-samd/blob/e6df2410613f4e940c2a84f2def3100bede2901a/cores/arduino/WString.cpp#L285

When adding '\0' to a String, the length will be incremented, but the NUL character is not written to the underlying buffer. The reason is that in concat() a strcpy is used which does nothing in this case. But the length is still incremented.

aentinger:
`\0` entries are currently not supported, however this may change if and when https://github.com/arduino/ArduinoCore-API/pull/97 is merged. Since those changes will be propagated via `ArduinoCore-API` integration into `ArduinoCore-samd` this issue can in fact be closed.

PR #371, opened 2018-11-07
Update WString

dav1901:
Prevent length from increasing when adding a NULL byte to a String. #279


aentinger:
Closing, see https://github.com/arduino/ArduinoCore-samd/issues/279#issuecomment-741658388 .

PR #372, opened 2018-11-08
Add isUARTEnable API

Rocketct:
Implemented API for status check: 
- Added API for check the UART status in Uart::flush()
- Added the API in SERCOM.h and SERCOM.cpp
 i have tested  the change with the following sketch:

```c++
void setup() {

}

void loop() {
  Serial1.begin(115200);
  Serial1.println("asdasdasdasfasdfafasdasdas_1");
  Serial1.end();
  Serial1.println("asdasfasfasfasfafsafasdfasfsdf_2");
  Serial1.flush();
}
```

without the change it stop in to the first time that flush is called.

cc/ @sandeepmistry @facchinm

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -108,6 +108,11 @@ void SERCOM::enableUART()
   while(sercom->USART.SYNCBUSY.bit.ENABLE);
 }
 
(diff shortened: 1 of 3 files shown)


sandeepmistry:
I'm going to close this for now, since we decided to take another approach inside the lib and don't need this change in the core.

PR #378, opened 2018-12-03
MKR boards: I2C to sercom2, SPI1 + Serial2 to sercom4

sandeepmistry:
This pull request changes the sercom used for I2C and SPI1/Serial2, so that sercom can be used for alternative purposes.

-> I2C: from sercom 0 to 2
-> SPI: from sercom 2 to 4

We'll have to be super careful and test the changes on all MKR board types.

@Rocketct if you have some time this week please test this out.


Rocketct:
@sandeepmistry tested both SPI and I2C by using the examples of each boards and the MKR ENV shield all goes fine, ok for me

sslupsky:
What is the purpose of this change?  The OP says "so that sercom can be used for alternative purposes".  What "alternative purposes" are you referring to?

sandeepmistry:
> What "alternative purposes" are you referring to?

Adding an additional SPI interface on pins A3 + A4 to use in conjunction with the new MKR RGB shield.

henklaak:
Late to the game, but the labels in the MKRWiFi1010V2.0 schematics (ABX00023-schematics.pdf) still reflect the old sercom numbers. Confused me no end (was there a board revision I had missed?), until I traced back this commit and the reasoning behind it.

Don't know where else I should report docs issues....

Screenshot from 2023-12-22 02-23-41

PR #379, opened 2018-12-20
USBDeviceClass::connected does not reflect actual connection state

chrulri:
Storing `FNUM` in an `uint8_t` would almost always reflect in `f` being different from `FNUM` even if `FNUM` did not change due to the information loss as `FNUM` is of type `uint16_t`.

The delay is required to give it time to change if the device is connected.
Although I'm not sure of the amount of delay required for it to change.

Tested with Arduino MKR WiFi 1010.

### DIFF
--- a/cores/arduino/USB/USBCore.cpp
+++ b/cores/arduino/USB/USBCore.cpp
@@ -520,8 +520,8 @@ void USBDeviceClass::stall(uint32_t ep)
 bool USBDeviceClass::connected()
 {
 	// Count frame numbers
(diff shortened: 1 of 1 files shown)


chrulri:
I did some measurements with my MKR WiFi 1010 and a delay below 900 microseconds may result in false negatives (`FNUM` did not change yet).
1 millisecond may be safe.
3 milliseconds are 100% safe according to my testing so far.

chrulri:
This PR has been superseded by https://github.com/arduino/ArduinoCore-samd/pull/664

PR #384, opened 2019-01-29
Add cores/util/atomic.h

WestfW:
Provide macros compatible with the avr-libc ATOMIC_BLOCK code.
Tested (only) by code inspection of the binary.


sslupsky:
Any chance this PR is going to be accepted soon?  Atomic block support is important for code that uses interrupts.

sslupsky:
See also this library:  https://github.com/wizard97/SimplyAtomic

PR #388, opened 2019-02-22
Bad i2c dev recovery

sampscl:
This fixes the issue I reported [to platformio](https://github.com/platformio/platformio-pkg-framework-arduinosam/issues/9). The downside is that `startTransmissionWIRE()` will immediately return an error on a multi-master I2C bus when another master currently owns the bus. Previous behavior was to wait indefinitely until the other master released the bus.

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -499,6 +499,19 @@ bool SERCOM::startTransmissionWIRE(uint8_t address, SercomWireReadWriteFlag flag
   // 7-bits address + 1-bits R/W
   address = (address << 0x1ul) | flag;
 
+  // If another master owns the bus or the last bus owner has not properly
+  // sent a stop, return failure early. This will prevent some misbehaved
+  // devices from deadlocking here at the cost of the caller being responsible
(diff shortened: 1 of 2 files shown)


facchinm (approved):

facchinm:
From the logic POW this patch looks fine; @sampscl do you have any test case that can help reproducing the issue in a specially crafted misbehaving multimaster setup?

sampscl:
I will work on it, maybe next week or the week after I'll have something?

Some history: I originally found this when I put an SMBUS battery we had laying around the lab onto my existing I2C bus so I could develop a smart battery interface. It turns out that SMBUS devices are multi-master and this particular one has some bug that causes it to inhibit the I2C stop condition in very rare circumstances (recently charged to 100%, drained a little bit maybe 97%, then start charging again and simultaneously queried for it's current battery state -- at least this was the most common condition when I saw the Arduino I2C deadlock). Check out my attached logic analyzer capture for reference.  I'll probably need to write some software to intentionally cause the issue so that it can be repeatedly observed outside my lab.

i2c-smbus-fail

PR #389, opened 2019-02-27
Changed POW_ON pin value on MKR1500's variant

Rocketct:
Changed POW_ON pin value on MKRNB1500's variant file, to allow the power on from MKRNB library.

cc/ @sandeepmistry as discussed here the change double check


sandeepmistry:
@facchinm me and @Rocketct are suggesting this change along with 
https://github.com/arduino-libraries/MKRNB/pull/8.

We noticed the power consumption was high when a blink sketch or the Arduino Low Power library examples were loaded on the board.

PR #392, opened 2019-03-20
[DO NOT MERGE] Update to CMSIS 5.5.0

sandeepmistry:
Depends on https://github.com/arduino/ArduinoModule-CMSIS/pull/3.

@cmaglie @facchinm what do you think about update CMSIS from 4.5 to 5.5?

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -64,8 +64,8 @@ compiler.S.extra_flags=
 compiler.ar.extra_flags=
 compiler.elf2hex.extra_flags=
 
-compiler.arm.cmsis.c.flags="-I{runtime.tools.CMSIS-4.5.0.path}/CMSIS/Include/" "-I{runtime.tools.CMSIS-Atmel-1.1.0.path}/CMSIS/Device/ATMEL/"
-compiler.arm.cmsis.ldflags="-L{runtime.tools.CMSIS-4.5.0.path}/CMSIS/Lib/GCC/" -larm_cortexM0l_math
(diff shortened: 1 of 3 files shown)


platform.txt
```diff
@@ -64,8 +64,8 @@ compiler.S.extra_flags=
 compiler.ar.extra_flags=
 compiler.elf2hex.extra_flags=
 
-compiler.arm.cmsis.c.flags="-I{runtime.tools.CMSIS-4.5.0.path}/CMSIS/Include/" "-I{runtime.tools.CMSIS-Atmel-1.1.0.path}/CMSIS/Device/ATMEL/"
-compiler.arm.cmsis.ldflags="-L{runtime.tools.CMSIS-4.5.0.path}/CMSIS/Lib/GCC/" -larm_cortexM0l_math
+compiler.arm.cmsis.c.flags="-I{runtime.tools.CMSIS-5.5.0.path}/CMSIS/Core/Include/" "-I{runtime.tools.CMSIS-5.5.0.path}/CMSIS/DSP/Include/" "-I{runtime.tools.CMSIS-5.5.0.path}/CMSIS/NN/Include/" "-I{runtime.tools.CMSIS-Atmel-1.1.0.path}/CMSIS/Device/ATMEL/"
```
sandeepmistry: I'm getting linker errors when trying to use NN functions, they are not pre-compiled like the DSP.

facchinm: @sandeepmistry do you think they are not precompiled because they use some features not commonly deployed across a microarchitecture family? In that case the only thing I can think about is turning them into Arduino libraries (which is bad, but not terrible, since CMSIS changes are in sync with the core)

PR #403, opened 2019-04-17
Reduce power consumption in DeepSleep

jesserod:
Commenting out INPUT pin initialization per https://github.com/adafruit/Adafruit_SleepyDog/issues/17#issuecomment-481798215

### DIFF
--- a/cores/arduino/wiring.c
+++ b/cores/arduino/wiring.c
@@ -79,12 +79,18 @@ void init( void )
 
   // Clock ADC/DAC for Analog
   PM->APBCMASK.reg |= PM_APBCMASK_ADC | PM_APBCMASK_DAC ;
-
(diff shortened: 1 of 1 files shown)


sslupsky:
I would like to propose a modification to this PR.

Would it be possible to instead of eliminating the initialization, use pinPeripheral() to do the initialization instead?  That is,

  // Setup all pins as configured in variant.c
  for (uint32_t ul = 0 ; ul < PINS_COUNT ; ul++ )
  {
    pinPeripheral( ul, g_APinDescription[ulPin].ulPinType );
  }

This has the benefit of configuring ALL the pins according to their defined use in variant.c.  All pins defined as "PIO_DIGITAL" are left in their default (reset) state.

PR #404, opened 2019-04-25
Removal of delay within USBDevice::ISRHandler

lxrobotics:
Previously a `delayMicroseconds(20);` was executed within `USBDevice::ISRHandler` in order to enable consistent triggering of bootloader upon firmware upload. This changes remove the need for the delay within the ISR.

### DIFF
--- a/cores/arduino/USB/USBCore.cpp
+++ b/cores/arduino/USB/USBCore.cpp
@@ -425,7 +425,7 @@ void USBDeviceClass::initEP(uint32_t ep, uint32_t config)
 
 		// Setup Control IN
 		usbd.epBank1SetSize(ep, 64);
-		usbd.epBank1SetAddress(ep, &udd_ep_in_cache_buffer[0]);
+		usbd.epBank1SetAddress(ep, &udd_ep_in_cache_buffer[ep]);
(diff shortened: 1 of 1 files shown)


sandeepmistry:
I've tested this on my Mac which runs macOS 10.14.4, with the following sketches:

* `VidorTestSketch`
* `VidorDrawLogo`
* `SerialPassthrough` with a wire between the TX and RX pins on the header

All looks good so far, please let me know if there are additional sketches that I can run.

lxrobotics:
Thank you very much for trying out with those sketches. I've used `VidorTestSketch` and `VidorDrawLogo` myself as well as the generic `Blink`. I don't believe any additional sketches need to be run, but maybe it's beneficial to test "aggressively" by uploading one sketch and then immediately follow with uploading the other sketch, vary the delay between completion of upload and trigger of next upload, ... . So far I did not manage to produce a hang-up with those methods.

lxrobotics:
I've tried the patch successfully with `VidorTestSketch` and `VidorDrawLogo` on both Ubuntu 18.04 LTS and Windows 10.

facchinm:
Works perfectly, merging!

PR #408, opened 2019-05-09
Better SPI SERCOM clock control

PaintYourDragon:
**DEPRECATES the M4 MAX_SPI compile-time setting** (and corresponding "Max SPI" menu) in favor of **run-time** SPI SERCOM clock options. MAX_SPI is still handled for any interim projects that required it, but it should not be used or needed going forward (it had problems in that it affected ALL SERCOMs, not just one SPI peripheral, and this caused problems with SD card access and likely other things as well).

Instead, the SPI library now provides the SPI.setClockSource(x) function, where x is one of SERCOM_CLOCK_SOURCE_48M (standard), SERCOM_CLOCK_SOURCE_100M (up to 50 MHz SPI) or SERCOM_CLOCK_SOURCE_FCPU (up to F_CPU/2 SPI). This can be called before or after SPI.begin(), and should also work on SPI1 and any “manually built” SERCOM SPI objects. This has no effect on M0 devices, but the function and enum are still present in that case to avoid confusion if novices copy-and-paste M4 code to an M0 project (SPI will just run at the standard 12 MHz max there).

ADDITIONALLY, the SPI library now provides the following functions:
- SPI.getDataRegister() returns a volatile uint32_t pointer which is the underlying SERCOM DATA register, required for DMA operations (but which was tricky to access before).
- SPI.getDMACID() returns the SERCOM DMAC TX identifier (an integer), again required by certain DMA operations.
- SPI.getSercomIndex() returns the underlying SERCOM index (an integer), e.g. 0 for SERCOM0, 1 for SERCOM1 and so forth (up to 5=SERCOM5 on most SAMD21s or 7=SERCOM7 on SAMD51). This is a catch-all should a project require other pointers or identifiers related to a SERCOM...a table of values can be used with this as an index.

(There's some underlying stuff going on in the core SERCOM.h and SERCOM.cpp files to support these changes, but this likely won't affect normal end users. At most they'll use the SPI library functions, and only in rare circumstances.)


PaintYourDragon:
That was SUPPOSED to go to adafruit:master, not arduino:master. Pardon the intrusion. :/

PR #412, opened 2019-05-26
Fixed an issue with interrupt pins

James-Aylward:
https://learn.adafruit.com/assets/47156
Fixed the discrepancy between the circuit playground diagram and the code. In the diagram, button A is labelled as an interrupt pin even though the code does not reflect this. This change fixes the issue.


facchinm:
LGTM!

PR #425, opened 2019-06-27
Fix array out of bounds in USB core

thepsr:
Fix array out of bounds problem that can randomly cause crashes.


facchinm:
Manually merged with the right title :wink: 
Thanks @thepsr for the contribution!

PR #427, opened 2019-07-02
Add RUNSTDBY flag to enable wake on I2C address match

sslupsky:
A Wire port can generate an interrupt on address match when sleeping to wake the mcu.  To do so requires that the I2C peripheral is enabled and has the "RUNSTDBY" bit in the CTRLA register set.

The current SAMD core does not have a method to set the RUNSTDBY bit.  This PR adds methods to the Wire and SERCOM libraries to support this.  The peripheral is disabled, the RUNSTDBY bit it configured and the peripheral is enabled again.

The name for the proposed Wire method was selected to be consistent with the only other instance of setting RUNSTDBY I could find which is located in the USBCore.

### DIFF
--- a/cores/arduino/SERCOM.cpp
+++ b/cores/arduino/SERCOM.cpp
@@ -413,6 +413,12 @@ void SERCOM::disableWIRE()
   }
 }
 
+void SERCOM::runstandbyWIRE(bool runstdby) {
+  disableWIRE();                                 //  RUNSTDBY bit is enable protected so disable the peripheral
+  sercom->I2CM.CTRLA.bit.RUNSTDBY = runstdby ;
+  enableWIRE();
+}
+
 void SERCOM::initSlaveWIRE( uint8_t ucAddress, bool enableGeneralCall )
 {
   // Initialize the peripheral clock and interruption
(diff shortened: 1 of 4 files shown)


cores/arduino/SERCOM.cpp
```diff
@@ -413,6 +413,12 @@ void SERCOM::disableWIRE()
   }
 }
 
+void SERCOM::runstandbyWIRE(bool runstdby) {
+  disableWIRE();                                 //  RUNSTDBY bit is enable protected so disable the peripheral
+  sercom->I2CM.CTRLA.bit.RUNSTDBY = runstdby ;
```
sandeepmistry: Is there any harm in setting this bit all the time on enable?

I'd prefer not to add platform specifics API's to the Wire lib.

sslupsky: There might be an increase in power consumption when RUNSTDBY is enabled.

sslupsky:
I derived another way to do this that only requires that the ArduinoCore-samd SERCOM and TWOWIRE classes are friends of the ArduinoLowPower class.

I will close this PR and open another one.

PR #430, opened 2019-07-24
fix inconsistent type usage between CDC.h and CDC.cpp

majbthrd:
Serial_::baud() is prototyped in CDC.h to return uint32_t, but then uses unsigned long in CDC.cpp

A variant of Serial_::begin() is prototyped in CDC.h to accept unsigned long as an argument, but then uses uint32_t in CDC.cpp

This patch standardizes both to use uint32_t.

### DIFF
--- a/cores/arduino/USB/CDC.cpp
+++ b/cores/arduino/USB/CDC.cpp
@@ -321,7 +321,7 @@ int32_t Serial_::readBreak() {
 	return ret;
 }
 
-unsigned long Serial_::baud() {
+uint32_t Serial_::baud() {
 	return _usbLineInfo.dwDTERate;
 }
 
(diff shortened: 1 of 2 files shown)


majbthrd:
Judging by other related classes, I'm inferring that the authors probably intended "unsigned long" as a standard baud rate type.  So, a revised patch will be submitted with standardization on unsigned long, rather than uint32_t.

matthijskooijman:
For future reference: You could have force pushed to the same branch to replace the contents of this PR with new commits, rather than closing and opening a new one (which is also fine of course, just FYI).

majbthrd:
Thanks.  It was my error; I tidied up too quickly and deleted my github fork of the code immediately after the initial pull request, so I had nothing left to force push to.

PR #432, opened 2019-07-25
Allows Arduino API users to handle custom USB events

PowerfulCat:


Pillar1989:
@facchinm This PR is also very useful for our software and we really hope to use this Easter egg on the samd21 platform

lxrobotics:
Thank you very much for creating this PR. However, since this is a very special use-case not concerning the typical Arduino user I'll unfortunately have to reject it.

PR #435, opened 2019-07-29
Add null pointer test to String destructor

jrowberg:
A rare but possible null pointer dereference occurs when a String object is destroyed after either a failed initialization or intentional invalidation. Adding a simple `if (buffer)` test before `free`ing it avoids this failure case.

### DIFF
--- a/cores/arduino/WString.cpp
+++ b/cores/arduino/WString.cpp
@@ -123,7 +123,7 @@ String::String(double value, unsigned char decimalPlaces)
 
 String::~String()
 {
-	free(buffer);
+	if (buffer) free(buffer);
 }
 
 /*********************************************/


matthijskooijman:
I do not think this is needed, libc defines `free(null)` as a noop. From the `free(3)` manpage:

>        The free() function frees the memory space pointed to by ptr, which must have been returned by a previous call to malloc(), calloc(), or realloc().  Otherwise, or if  free(ptr)  has  already  been
       called before, undefined behavior occurs.  If ptr is NULL, no operation is performed.

I see this PR was already reverted by @lxrobotics, possibly for this reason?

lxrobotics:
@matthijskooijman you're right, according to spec `free` does nothing if the provided argument is 0.

PR #436, opened 2019-08-02
Add virtual base for method "ready" in Client

TechNyquist:
Hello everyone,

I'm shooting in the dark as of what could imply this but, what about adding the virtual base for "ready" method in Client class?
I've tested it and works seamlessly with my current code, meanwhile it allows for polymorphic abstraction when building a class that should work asynchronously with all the possible clients (GPRS, WiFi, etc.) polimorphically.

The "ready()" method is mandatory when working asynchronously (a great feature when dealing with whatchdog), but Client base class lacks its declaration. Because of this one should wrap the whole in a class just to add the "ready" method, or reimplement over and over almost the same code for every possible transport. Having the base instead would allow one to define the asynch code once for all.

I added a default body instead of making it pure virtual just to don't oblige to implement it even where not required (ie. MqttClient complained about this).

Again, I don't know if this clashes with your class tree logic, but I believe it would be seamless and yet useful when working through advanced c++ ways (so not for normal use).

### DIFF
--- a/cores/arduino/Client.h
+++ b/cores/arduino/Client.h
@@ -37,6 +37,7 @@ public:
   virtual void flush() = 0;
   virtual void stop() = 0;
   virtual uint8_t connected() = 0;
+  virtual int ready() { return 1; }
   virtual operator bool() = 0;
 protected:
   uint8_t* rawIPAddress(IPAddress& addr) { return addr.raw_address(); };


TechNyquist:
Uhm, anybody...?

PaulStoffregen:
Many explain what you intend this ready() function to actually do?  Everything you've written so far sounds like you assume we already know.

Maybe also explain how this would be different from availableForWrite() which exists in the HardwareSerial class.

sslupsky:
Agree with Paul.  What is the purpose for this?

PaulStoffregen:
Might also be worth mentioning availableForWrite() is indeed part of the Print class, which Stream and Client inherit.

https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/Print.h#L60
https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/Print.h#L63
https://github.com/arduino/ArduinoCore-API/blob/master/api/Print.h#L59

If the intended purpose of ready() is the same as availableForWrite(), then this issue can probably be considered solved, right?

TechNyquist:
Please let me clear up, so to get synch.
In order to minimize codebase to maintain I wrapped "Client" pointer in my own class, this way I can enforce advanced logics (such as retry strategies and so on) on clients behaviour, regardless of their specific type (GSMClient, WiFiClient, ...whatever).

Simple example:
```
class ClientWrapper
{
   Client *m_Client;

   public:
      ClientWrapper(Client *client)
         :m_Client(client)
      {
      }

      void connect(...)
      {
         if( m_Client->connect() )
         {
            ...
         }
         else
         {
            ...
          }
      }

      void disconnect()
      {
         m_Client->disconnect();
      }

      ...
}

void setup()
{
   ...
   GSMClient client;
   ClientWrapper w(&client);
   if( w.connect() )
   {
      ...
   }
   ...
}
```
This is very simplified but goes straight to the point: ClientWrapper can wrap every client inherited from Client, so it uses the same logic for GSMClient, so for WiFiClient and so on.
All is working great.

But.
In order to properly handle watchdog timeout I cannot go for synchronous operation (such as "connect" ie) because it might stop the run longer than WDT and cause unnecessary reset. Because of this I require to work asynchronously. When working asynchronously the "ready" method is used after every operation to know its outcome.
Here comes the problem: 
- "ready" is first defined in "GSMClient" ([here](https://github.com/arduino-libraries/MKRGSM/blob/b13ad329093a35ee33687de90639c4561a9d6353/src/GSMClient.h#L47))
- but "ready" virtual declaration is not present in Client class (https://github.com/lstoll/arduino-libraries/blob/master/Ethernet/Client.h), so adding it with some default successful value would abstract away the whole, and is how polymorphism should be meant to work IMHO.

Right now to make it work I have to explicitly cast to GSMClient (I'm using only GSM for now), but of course this is a breaking workaround.

GOTCHA: I notice now that WiFiClient is not providing such "ready" method and asynch at all by now... did I walk too far expecting asynch management everywhere?
If you think that this PR asks for something that would required much deeper refactoring and logic change to apply, feel free to reject it.

PR #442, opened 2019-08-15
Fix requestFrom() return type for 16-bit lengths

jrowberg:
The latest changes to the SAMD core create a 256-byte templated ring buffer for received data in the `Wire` class. However, the `requestFrom` methods only support a `uint8_t` return type, which means they report `0` for the number of bytes received after a 256-byte request (I found this out the hard way). This fix changes `uint8_t` to `size_t` to match the `quantity` argument data type and the `byteRead` variable data type, which is the one actually used to provide the return value.

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -64,7 +64,7 @@ void TwoWire::end() {
   sercom->disableWIRE();
 }
 
-uint8_t TwoWire::requestFrom(uint8_t address, size_t quantity, bool stopBit)
+size_t TwoWire::requestFrom(uint8_t address, size_t quantity, bool stopBit)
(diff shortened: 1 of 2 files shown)


sandeepmistry:
@facchinm @cmaglie I think we should go with this, but it impacts all the cores and `ArduinoCore-API`: https://github.com/arduino/ArduinoCore-API/blob/master/api/HardwareI2C.h

facchinm:
It looks reasonable indeed; no code should break due to automatic promotion but we need to change the non virtual implementation in mbed and megaAVR core if we apply it to API repo.

aentinger:
Depends on https://github.com/arduino/ArduinoCore-API/pull/132.

PR #447, opened 2019-09-03
Bugfix - waiting for first ADC read to complete after ADC enable.

### LINKED ISSUE #446: analogRead() provides incorrect results on the SAMD21 core.
Consider the following sketch:

```
#define Debug Serial

void setup() {
  analogReadResolution(12);
  pinMode(A0, INPUT);
  Debug.begin(9600);
  while (!Debug);
}

void loop() {
  int a = analogRead(A0);
  float v = 3.3*((float)a)/4095;

  Debug.print(a);
  Debug.print(" ");
  Debug.print(v,3);
  Debug.println();

  delay(1000);
}
```

If you provide a voltage on pin A0, then results in `v` are typically offset by +35mV (sometimes a little less, sometimes a little more). 

There is likely a bug in the Arduino code of the `analogRead()` function, as detailed here:
https://www.omzlo.com/articles/your-arduino-samd21-adc-is-wrong-did-you-notice

lxrobotics:
This PR fixes #446.

### DIFF
--- a/cores/arduino/wiring_analog.c
+++ b/cores/arduino/wiring_analog.c
@@ -165,6 +165,9 @@ uint32_t analogRead(uint32_t pin)
   syncADC();
   ADC->SWTRIG.bit.START = 1;
 
+  // Waiting for the 1st conversion to complete
+  while (ADC->INTFLAG.bit.RESRDY == 0);
+
   // Clear the Data Ready flag
   ADC->INTFLAG.reg = ADC_INTFLAG_RESRDY;
 


amezins:
I had a similar problem on an AD674  used in a project over 20 years ago. All clone manufacturers duplicated the problem of the original part that failed to include a HW reset line to initialize the internal registers that weren't reset by the power up cycle.  I "fixed" it by adding a dummy conversion start command to the ADC driver's power up initialization. Each of my drivers had a vector for power up initialization that were enumerated by the "main" power up initialization for the drivers.

Muplex:
Hello,
I have last build installed 1.8.3 but these bug still happend. 
I measure 30 to 40 steps (12 bit resolution) when analogpin is grounded (should be near to 0) 
Can you explain me how to fix it ?

Thank you

lxrobotics:
That's because no new version (next would be `1.8.4`) has been released. @facchinm What do you think about releasing `1.8.4`?

facchinm:
@lxrobotics today or tomorrow :wink:

PR #448, opened 2019-09-12
fix charging safety timer issue

Rocketct:
This changes allow to avoid freeze module's freeze due to fault for charging safety timer expiration for boards where bq24195 is used.

cc/ @sandeepmistry @facchinm the test still running after 22 hours, wait another day or two in order to see if others faults are raised and after we could merge


Rocketct:
this change are, Dangerous in MKRNB closed

PR #449, opened 2019-09-16
Fixing infinite hang within 'noTone' if it is called before 'tone'

### LINKED ISSUE #326: Bug when calling noTone() before tone
If you happen to call noTone() before ever calling tone(), the sketch hangs at line 45 of tone:
  WAIT_TC16_REGS_SYNC(TCx)
because the timer has never been configured.

lxrobotics:
`tone` need to run at least once in order to enable GCLK for the timers used for the tone-functionality. If `noTone` is called without ever calling `tone` before then `WAIT_TC16_REGS_SYNC(TCx)` will wait infinitely. The variable `firstTimeRunning` is set the 1st time `tone` is set so it can be used to detect wether or not `tone` has been called before. Fixes #326 .

### DIFF
--- a/cores/arduino/Tone.cpp
+++ b/cores/arduino/Tone.cpp
@@ -152,9 +152,19 @@ void tone (uint32_t outputPin, uint32_t frequency, uint32_t duration)
 
 void noTone (uint32_t outputPin)
 {
-  resetTC(TONE_TC);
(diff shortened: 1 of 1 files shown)


facchinm:
While not amused by the name of `firstTimeRunning` variable, the fix is fine :slightly_smiling_face:

PR #456, opened 2019-09-19
Remove Genuino Occurrences

mastrolinux:


boards.txt
```diff
@@ -55,15 +55,15 @@ arduino_zero_native.pid.2=0x824d
 # If the board is a 2341:824d use 2341:824d for build and set other parameters as well
 arduino_zero_native.vid.2.build.vid=0x2341
 arduino_zero_native.vid.2.build.pid=0x824d
-arduino_zero_native.vid.2.build.usb_product="Genuino Zero"
+arduino_zero_native.vid.2.build.usb_product="Arduino Zero"
 arduino_zero_native.vid.2.bootloader.file=zero/samd21_sam_ba_genuino.bin
```
cmaglie: The meaning of this piece:

```
arduino_zero_native.vid.2=0x2341
arduino_zero_native.pid.2=0x824d
# If the board is a 2341:824d use 2341:824d for build and set other parameters as well
arduino_zero_native.vid.2.build.vid=0x2341
arduino_zero_native.vid.2.build.pid=0x824d
arduino_zero_native.vid.2.build.usb_product="Genuino Zero"
arduino_zero_native.vid.2.bootloader.file=zero/samd21_sam_ba_genuino.bin
```

can be summed up as:

- If you find a board with vid/pid `2341:824d` then
- build the sketch with vid/pid `2341:824d` and use usb_product `Genuino Zero` and (possibily) burn genuino bootloader.

but since there is no more need to keep `Genuino` we can just remove the extra section and leave just:

```
 arduino_zero_native.vid.2=0x2341
 arduino_zero_native.pid.2=0x824d
```

so the board is "detected" as an "Arduino Zero" and the PID 824d will be replaced with the Arduino PID 804d at the next upload.

bootloaders/zero/build_all_bootloaders.sh
```diff
@@ -2,14 +2,9 @@
...
 
-BOARD_ID=genuino_zero NAME=samd21_sam_ba_genuino make clean all
-
 BOARD_ID=arduino_mkr1000 NAME=samd21_sam_ba_arduino_mkr1000 make clean all
 mv -v samd21_sam_ba_arduino_mkr1000.* ../mkr1000/
 
-BOARD_ID=genuino_mkr1000 NAME=samd21_sam_ba_genuino_mkr1000 make clean all
-mv -v samd21_sam_ba_genuino_mkr1000.* ../mkr1000/
-
 BOARD_ID=arduino_mkrzero NAME=samd21_sam_ba_arduino_mkrzero make clean all
```
cmaglie: We may want to remove also the binaries:
```
mkr1000/samd21_sam_ba_genuino_mkr1000.bin
zero/samd21_sam_ba_genuino.bin
```

matthijskooijman:
Is this a good idea? I thought about it recently, but realized that people will (I think?) have boards lying around that are labeled "Genuino" and might now be confused about how to use them? OTOH, people with Arduino boards might also be confused by the Genuino mention, so I guess there's a balance there...

mastrolinux:
The idea was to create the minimum amount of confusion, given the fact Genuino boards are numerically much less than Arduino boards, we decided to remove the Genuino word while still supporting the boards.

PR #458, opened 2019-09-30
Added printf to Print class to have Serial.printf() functionality

beegee-tokyo:
Adafruits version of the SAMD Arduino framework has Print::printf function. As the SAMD MCUs are not as memory limited than the AVR 8 bit MCUs I think it will not hurt to add it.

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -183,6 +183,16 @@ size_t Print::println(const Printable& x)
   return n;
 }
 
+void Print::printf(const char format[], ...)
+{
+  char buf[PRINTF_BUF];
+  va_list ap;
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Thanks for your contribution. However, I'm closing this, since:
 - Changes to Print.h should be made through the arch-independent https://github.com/arduino/ArduinoCore-API repo in the future
 - There is already a similar PR there: https://github.com/arduino/ArduinoCore-API/pull/28
 - So far, the Arduino team has specifically decided against including printf, see https://github.com/arduino/ArduinoCore-API/issues/32 and https://github.com/arduino/ArduinoCore-API/pull/28#issuecomment-468471093

PR #469, opened 2019-11-20
SNU/SFU - Second stage bootloader for WiFi and Serial flash based OTA

aentinger:
**SNU** is a second stage bootloader for SAMD boards with WiFi NINA module (MKR WIFI 1010 and MKR VIDOR 4000). The SNU is loaded by the ATSAMD21G18 bootloader and in turn loads the firmware image stored in WiFi NINA W102 module.

Memory map of ATSAMD21G18 flash space with bootloader, SNU and application:
at90samd21g18-ota-flash-nina-storage

**SFU** is a second stage bootloader for SAMD boards when an external serial flash connected. Since the [SerialFlash](https://github.com/PaulStoffregen/SerialFlash) library is used all serial flash memories supported by this library are supported. Memory map is identical with the SNU memory map, but instead of the SNU binary the SFU binary is included in the two consecutive flash sections after the bootloader.


aentinger:
I'm closing this PR because SFU/SNU were separated in two different PRs and merged separately (#512, #513).

PR #474, opened 2019-12-08
Update I2S.cpp WDT patch

skorpsim:
enableClock() in I2S.cpp does not set DIVSEL bit but requires DIVSEL to be 0.
All of the WDT's i have tested so far set DIVSEL to 1.
This leads to incompatibilities. For example no sound with ArduinoSound and WDT


CptHolzschnauz:
This PR solves the problem with the I2S / WDT incompatibility.

Why was it removed from the merge list? What is the issue with this PR?

Issue #462, opened 2019-10-10
Fix Nano 33 IoT variant comments


facchinm:
https://github.com/arduino/ArduinoCore-samd/blob/master/variants/nano_33_iot/variant.cpp#L19 need to be adapted to the actual pinout (following the correct table below)

PR #475, opened 2019-12-09
Modified the nano 33 Iot variant

Rocketct:
Adapted the MCU table description with the actual PINOUT

cc:/ @facchinm could you double check?

resolve: https://github.com/arduino/ArduinoCore-samd/issues/462


facchinm:
LGTM!

PR #509, opened 2020-04-02
Set USB SRAM QoS to sensitive latency

sameer:
By default this is `00` for both DQOS and CQOS, which means that USB is treated as a background operation and there is at least a one clock-cycle SRAM access latency. This reduces the SRAM clock-cycle latency for USB by setting it to `10`, giving it higher memory access priority.

This is done by [Adafruit](https://github.com/adafruit/uf2-samdx1/blob/master/src/startup_samd21.c#L254) and by the [Rust SAMD21 support crate](https://github.com/atsamd-rs/atsamd/blob/master/hal/src/samd21/usb/bus.rs#L690) (albeit at critical sensitivity) to improve USB performance and ensure correct behavior.
[Relevant datasheet section for QOSCTRL](http://ww1.microchip.com/downloads/en/DeviceDoc/SAM_D21_DA1_Family_DataSheet_DS40001882F.pdf#_OPENTOPIC_TOC_PROCESSING_d138e244709)
[Relevant datasheet explanation for QoS](http://ww1.microchip.com/downloads/en/DeviceDoc/SAM_D21_DA1_Family_DataSheet_DS40001882F.pdf#_OPENTOPIC_TOC_PROCESSING_d138e244709)

### DIFF
--- a/cores/arduino/USB/SAMD21_USBDevice.h
+++ b/cores/arduino/USB/SAMD21_USBDevice.h
@@ -50,6 +50,10 @@ public:
 	inline void noRunInStandby() { usb.CTRLA.bit.RUNSTDBY = 0; }
 	inline void wakeupHost()     { usb.CTRLB.bit.UPRSM = 1; }
 
+	// USB QoS
+	inline void setDataSensitiveQoS() { usb.QOSCTRL.bit.DQOS = 2; }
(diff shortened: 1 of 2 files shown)


facchinm:
LGTM! Merging, thanks!

PR #511, opened 2020-04-07
Adds to Wstring

shawayush:
It gives a solution to issue #279 . But somehow I feel, a little more modification is needed, so anyone can suggest something to it, it will really be helpful.
Thank you


aentinger:
Closing, see https://github.com/arduino/ArduinoCore-samd/issues/279#issuecomment-741658388 .

PR #513, opened 2020-04-15
SFU - Second Stage Bootloader for MKR Boards with MKRMEM Shield

aentinger:
SFU is a second stage bootloader for MKR boards with MKRMEM shield. The SFU is loaded by the ATSAMD21G18 bootloader and in turn loads the firmware image stored on the MKRMEM shield.

**Requires:** Arduino MKR board + Arduino MKR MEM Shield

**How-to-use:**
* Include <SFU.h> at the top of your sketch
```C
#include <SFU.h>
```
* Export the binary via Menu **Sketch -> Export Compiled Binary**
* Convert sketch to C byte array file `Binary.h`
```bash
cat MY_SKETCH.bin | xxd --include > Binary.h
```
* Copy `Binary.h` to [SFU/examples/SFU_LoadBinary](https://github.com/arduino/ArduinoCore-samd/blob/SFU/libraries/SFU/examples/SFU_LoadBinary)
* Use sketch `SFU_LoadBinary` to download binary update file to MKRMEM shield
* Upload your SFU-enabled sketch - the update is performed automatically and then your sketch is started.


aentinger:
Successfully tested with `MKR WiFi 1010` + `MKR MEM Shield`.

PR #516, opened 2020-04-27
Changing "NONA WLAN" in "Nano 33 IoT"

luigigubello:
Changing "NONA WLAN" in "Nano 33 IoT"


aentinger (approved):
It looks good to me, the only line that is actually using the define `STRING_PRODUCT` is [this one](https://github.com/arduino/ArduinoCore-samd/blob/master/bootloaders/zero/sam_ba_usb.c#L220). @facchinm I'm wondering if accepting this change would have any undesirable side-effects?

PR #546, opened 2020-07-15
Add SSU LZSS HTTP OTA example

manchoz:
Add example for SSU OTA via HTTP with an lzss-compressed file.

@aentinger here we ago


manchoz:
I think we need we need a few more 3/4-letters acronyms for the title of next PRs 🤣

aentinger:
Well, the `SSU` for MKR NB 1500 will be `SBU` :stuck_out_tongue: @giulcioffi is working on that one already :wink:

PR #549, opened 2020-07-16
Add Second Stage Bootloader (SBU) for MKRNB 1500

giulcioffi:
SBU is a Second Stage bootloader for board MKRNB 1500 which uses Sara-R410M module. It follows the implementation of SSU for MKRGSM 1400, but using the power-on/off & reset sequences proper of Sara-R410M module ([MKRNB/#47](https://github.com/arduino-libraries/MKRNB/issues/47)).
It exploits the File System APIs in library MKRNB to perform file operations from the module.
An example is provided to test how a file is loaded into the Sara module and then retrieved at the boot and flashed to be executed.
1. Compile SBUBoot
2. Go to /SBU/extras/SBUBoot and run ./build.sh
3. Upload the example SBU_Usage.ino and follow the instructions written there.
4. Upload the example SBU_LoadBinary.ino. 
     - the binary of SBU_Usage will be loaded onto the Sara Module in a file called UPDATE.BIN
     - after a successful download, the file UPDATE.OK will be created and stored in the module
     - a system reboot is triggered
     - SBU checks if UPDATE.OK is present. If it finds it, it starts reading UPDATE.BIN and writing it into the flash by chunks of 512 bytes.
     - the flashed binary of SBU_Usage.ino is executed


libraries/SBU/extras/SBUBoot/SBUBoot.ino
```diff
@@ -0,0 +1,108 @@
...
+{
+  init();
+
+  __libc_init_array();
+
+  delay(1);
+
+  constexpr size_t blockSize = 512;
+  fileUtils.begin();
+  Serial1.begin(115200);
```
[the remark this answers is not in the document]
giulcioffi: Now they should be all gone :)

libraries/SBU/extras/SBUBoot/SBUBoot.ino
```diff
@@ -0,0 +1,101 @@
...
+        digitalWrite(LED_BUILTIN, LOW);
+        read_bytes = fileUtils.readBlock(UPDATE_FILE_NAME, tot_bytes + SBU_SIZE + SBU_START, spare_bytes, block);
+        digitalWrite(LED_BUILTIN, HIGH);
+        mcu_flash.write((void*)flash_address, block, read_bytes);
+        flash_address += read_bytes;
+      }
+      update_success = true;
+    }
+    if (update_success) {
+      fileUtils.deleteFiles();
```
aentinger, line 86: Question ;) Will `fileUtils.deleteFiles()` any files? Like also ones that the user is storing on the MKR NB 1500 modem? That's quite a realistic possibility given the fact that we provide that capability to the user via `NBFileUtils`.

libraries/SBU/extras/SBUBoot/SBUBoot.ino
```diff
@@ -0,0 +1,102 @@
...
+  // Try to update only if update file
+  // has been download successfully.
+
+  if (fileUtils.existFile(CHECK_FILE_NAME)) {
+    uint32_t updateSize = fileUtils.listFile(UPDATE_FILE_NAME);
+    uint32_t tot_bytes = 0;
+    uint32_t read_bytes = 0;
+
+    if (updateSize > SBU_SIZE) {
+      updateSize = updateSize - SBU_SIZE - SBU_START;
```
aentinger, line 58: Good Morning @giulcioffi ☕ 👋 I don't think this is quite right. The reason for this is that the first 0x2000 bytes of the MCU flash are occupied by the bootloader (the default bootloader, not the SBU) but those 0x2000 bytes of default bootloader are not included in the binary generated when compiling your Arduino sketch. Therefore I think it very likely that with the code like this you're skipping the last 0x2000 bytes of the actual application. 
```diff
-updateSize = updateSize - SBU_SIZE - SBU_START;
+updateSize = updateSize - SBU_SIZE;
```
For reference compare to [SSU](https://github.com/arduino/ArduinoCore-samd/blob/master/libraries/SSU/extras/SSUBoot/SSUBoot.ino#L110).

variants/mkrnb1500/variant.cpp
```diff
@@ -254,7 +254,7 @@ void initVariant() {
 
   // put SARA modem in reset on start to conserve power if it's not used
   pinMode(SARA_RESETN, OUTPUT);
-  digitalWrite(SARA_RESETN, HIGH);
+  digitalWrite(SARA_RESETN, LOW);
```
aentinger, line 257: Hi @giulcioffi :coffee: So we are still putting the SARA in reset here. Is that still necessary given the changes in https://github.com/arduino-libraries/MKRNB/pull/55 ? Cheers, Alex - Otherwise great PR ;) let's hash this one out in Slack and then we should be able to merge it in.

aentinger: All cleared ;)

PR #554, opened 2020-08-17
SNU check min required version of nina-fw

aentinger:
FYI @eclipse1985


eclipse1985:
> FYI @eclipse1985

👍

PR #569, opened 2020-11-02
Refactor and improve PMIC configuration

kpfleming:
* Add PMIC.h and PMIC.cpp for general PMIC support in variants.

* Add BQ24195.h with register addresses and definitions for
BQ24195/BQ24195L devices.

* Add setupPMIC function to provide consistent initialization,
including:
    * Disable I2C watchdog timer to keep device in 'host' mode.
    * Disable charging when no battery present.
    * Disable charging system safety timer when no battery present.
    * Disable battery FET when no battery present.
    * Enable USB D+/D- on boards where they are connected to PMIC.
    * Enable 2A input current limit and 3.88V input voltage limit
       to enable high-speed battery charging.

* Update all variants with BQ24195 devices to use setupPMIC function.

Results:

* STAT LED no longer blinks on devices without batteries (was caused
by charging system safety timer expiration).

* Charging is actually disabled on boards without batteries (previously
it was left as the default, which is 'enabled' in these PMICs).

* Charging speed is dramatically increased (assumes 2A USB power supply
    is used to power board; MKR GSM 1400 previously assumed 3A power supply).

Testing:

* Has been tested on MKR WiFi 1010 with no battery and with Adafruit 2000mAh LiPo battery.

### DIFF
--- a/cores/arduino/PMIC.cpp
+++ b/cores/arduino/PMIC.cpp
@@ -1,0 +1,79 @@
+/*
+  PMIC.cpp - initialization of Power Management ICs
+  Copyright (c) 2020 Kevin P. Fleming.  All right reserved.
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#include "Arduino.h"
+
+#ifdef USE_BQ24195L_PMIC
+
+#include "PMIC/BQ24195.h"
+
+#include "wiring_private.h"
+
+void setupPMIC(SERCOM& sercom, bool batteryPresent, bool USBDetect) {
+  BQ24195_REG00 reg00;
+  reg00.IINLIM = 0b110; // input current limit 2A
+  reg00.VINDPM = 0b0000; // input voltage limit 3.88V
+  reg00.EN_HIZ = 0b0; // disable
+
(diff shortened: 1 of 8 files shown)


kpfleming:
There is at least one open item: two of the variants had code to set the input voltage limit and input current limit (and not to the same values), the others did not set these limits. A comparison of the schematics for all the variants will be required to determine if these should indeed be set for all variants, and to which values.

kpfleming:
This has now been tested through multiple cycles on a MKR WiFi 1010 with a 2000mAh battery; from full charge through full discharge and back to full charge again.

ronnyek:
I am interested in seeing this, as I'd like to take advantage of battery stuff. What does it take to get this merged in?

ninouing:
Hi @kpfleming, I was planning to use your code for increasing the charging current of an MKR WiFi 1010 to 2A as you propose.

I have one concern. If I look at the schematic of the MKR WiFi 1010 provided by Arduino I see that in series with the battery connector there is a fuse (MF-FSMF050) which from datasheet can hold a max of 0.5A. As you say that you tested this code without issues, can you please help me understand what am I missing? can the current go above the 0.5A? wouldn't the fuse burn out?

Thanks in advance

kpfleming:
It's been a while since I worked on this, but I believe that the '2A Input Current Limit" referred to is the input power from the USB connector; it's not the charging current sent to the battery. The PMIC handles battery-charging current itself, based on the battery's response, and I suspect that the battery will not accept more than 0.5A of charging current so the fuse is properly sized.

ninouing:
Hi kpfleming, thank you I appreciate your feedback.  I now understand.

I looked at the BQ24195L datasheet and indeed you are right, there are two settings. One is for the current drawn from USB (which you modify), the other is for the max current sent to the battery. This last value is 2A by default but the Arduino changes this register to 0.5A, thus the battery, even if you draw 2A from USB, is recharged at max 0.5A. 

This means that maybe the statement "enable high-speed battery charging" is a bit misleading, because it only solves the issue of the reduced input current from USB (0.1 A due to D+/D- not connected to BQ24195L).  You could set up a 0.5A current from USB instead of 2A and you would get the same charge time. 

Otherwise, you could change the register for the charging current, but then the fuse would burn...

In my case this is important because I'm planning to use a large capacity battery (7800 mA/h), which would require about 15h recharge time at 0.5A, compared to about 4h at 2A. 

I guess I will need to replace the fuse...

Thanks again!

kpfleming:
That's almost accurate I believe: if you set the input current limit to 0.5A, then the battery could be charged at a maximum of 0.5A minus the current being drawn by the Arduino and any devices attached to it. That's not typically a large amount of current, but it's non-zero, so should be factored into the equation.

Issue #576, opened 2020-12-15
 Problems with the new Arduino Samd Board 1.8.10 (32bit ARM Cortex-M0+) Version


CptHolzschnauz:
After upgrading to 1.8.10 the compiler cant't find #include <avr/pgmspace.h> anymore (File not found)
The whole controller was throttled down to maybe 5% of it's norma speed, everything was working like in super slow motion.
After including the pgmspace.h manually nothing changed, eveything in slow motion.

Then i downgraded to 1.8.9 and everything was back to normal...strange.
I can't give additonal infos, the compiler went trough normal.

facchinm:
Hi @CptHolzschnauz ,
this is very concerting, thanks for letting us know.
Can you post your sketch (or a snippet exposing the erratic behaviour) and other info like board in use? Thanks

CptHolzschnauz:
Hello, Master of the Core!
I am using a MKR 1400 and my sketch is about 3500 lines. After upgrading to 1.8.10 the compiler can't find  the file with #include <avr/pgmspace.h> anymore. So i included it manually (copy the .h and .cpp into my sketch folder and
#include pgmspace.h). After that  everything worked like normal (Compiling, Uploading, the Sketch) but the the whole Arduino was working in super slow motion (Like CPU throttled down to 5 %). I did not spend to much time in examination because after a simple downgrad to 1.8.9 everything went back to normal.
The slow motion i saw was mainly in displaying graphics from PROGMEM and with the Adafruit GFX Lib.
I think there is a link between these two events (pgmspace not found, slowdown, NVRAM, Flash ) and something is wrong there.
Thats all i can say for the moment. Tell me if i can help furtherwise, i have a black magic probe here but i need a hint.
Thx for your great work anyway!
With best regards
PS: Please wire the I2S an the MKR1500 like you did on the MKR1400 otherwise the board can't make voicecalls what i desperately need...and they are about to shut down 2G and 3G so the MKR1400 won't work anymore...

facchinm:
Ok, let's try to put together some info:
* SPI speed calculation has hanged between 1.8.9 and 1.8.10 . In fact, some frequencies that were not reached before are now ok. This could be a problem with some hardware.
From `Adafruit_ILI9341` driver, for example, default SPI frequency is 24000000 . With core 1.8.9 we are sending a 12MHz clock, while 1.8.10 produces the correct frequency.
[...]

PR #579, opened 2020-12-17
Limit max SPI clock frequency to F_CPU/2

facchinm:
Fixes https://github.com/arduino/ArduinoCore-samd/issues/576

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -93,8 +93,13 @@ void SPIClass::config(SPISettings settings)
     this->settings = settings;
     _p_sercom->disableSPI();
 
-  _p_sercom->initSPI(_padTx, _padRx, SPI_CHAR_SIZE_8_BITS, getBitOrder(settings));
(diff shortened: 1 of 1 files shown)


aentinger (approved):
LGTM :+1:

matthijskooijman:
What's the exact problem here? Does something overflow when the SPI speed > F_CPU/2? Where does this limit come from exactly? IMHO the code and commit message could use a bit more details for future reference.

facchinm:
@matthijskooijman agree, this is the direct porting of https://github.com/arduino/ArduinoCore-samd/blob/1.8.9/libraries/SPI/SPI.h#L40-L46 (which was not merged during API migration).
I'll add some info to the commit message

giulcioffi (approved):

matthijskooijman:
Thanks for updating the commit message, looks good to me. I do see that one testcase failed, was that already failing on master?

facchinm:
It's failing due to rate limiting :sweat_smile: Everything should be fine

Issue #580, opened 2020-12-16
Error while writing flash memory - Arduino Nano 33 IoT


tessi:
Hi :)

I just received an original Arduino nano 33 IoT and wanted to test the wifi. The firmware check told me I need to update, and I followed the instructions to update the firmware. Instead of the nice success-message, I received the following

<img width="933" alt="Screenshot 2020-12-16 at 12 56 45" src="https://user-images.githubusercontent.com/206108/102354971-864bbc80-3fab-11eb-9219-5ab1c974dcf6.png">

Used versions:

* Arduino GUI: 1.8.13
* WiFiNINA: 1.8.0
* WiFi101-FirmwareUpdater-Plugin: the bundled version and [version 0.10.12](https://github.com/arduino/WiFi101-FirmwareUpdater-Plugin/releases/tag/v0.10.12)
* macOS catalina (10.15.7)

The flashing failed multiple times even after restarting, reconnecting, re-flashing things multiple times.
The firmware-check now reports that it can't connect to the wifi module (it used to work before):

```
14:36:58.459 -> WiFiNINA firmware check.
14:36:58.459 -> 
14:36:59.172 -> Communication with WiFi module failed!
```

I am at a loss on how to continue debugging this. Do you have any hints?
How could I resolve this issue and is it even something I should report here?

Note: I would be happy to ignore this issue, but now I can't use wifi from normal sketches anymore, which is a little frustrating.

Damiasroca:
> Hi @tessi and @Damiasroca ,
> it looks like the issue has been caused by samd core 1.8.10 release. I'm moving the issue on that repo to ease keeping track of it. In the meantime, please revert to samd core 1.8.9 if you need to update the firmware. Will keep you posted as soon as the issue is solved 🙂

Downgrading indeed did the trick, Up and running again. Thank you

tessi:
<img width="373" alt="Screenshot 2020-12-17 at 20 41 52" src="https://user-images.githubusercontent.com/206108/102535571-af05ac00-40a8-11eb-95ec-e957229e8b18.png">

Downgrading indeed worked, thank you for the super helpful and quick response @facchinm  and the fix in progress @aentinger  ❤️

PR #582, opened 2020-12-18
Provide irq safe implementation of RingBuffer

facchinm:
Alternative way to fix https://github.com/arduino/ArduinoCore-samd/issues/580 without restoring the buggy RingBuffer implementation from API

### DIFF
--- a/cores/arduino/SafeRingBuffer.h
+++ b/cores/arduino/SafeRingBuffer.h
@@ -1,0 +1,56 @@
+/*
+  Copyright (c) 2020 Arduino.  All right reserved.
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


aentinger (approved):
LGTM :+1: Thank you @facchinm :rocket:

matthijskooijman (approved):
Code also looks good to me.

I wonder: Is this class something that should live in -API? I can imagine that this code is moved to API, and then individual cores make available a "sync.h" that -API can use to get platform-specific "synchronized" behaviour. As an added bonus, this would mean that `synchronized` can become a nice and portable way to disable interrupts (though I wonder if `synchronized` is the perfect name, but I guess it's modeled after Java, right?).

facchinm:
Yes, it's modeled over Java and could have a more embedd-ish name if we decide to port it to -API :slightly_smiling_face: . 
The problem in doing so is that we must ask for a `sync.h` or similar to be present in the core implementation, and given the include order (`Arduino.h` depends on `ArduinoAPI.h`) so we can't easily do feature discovery via defines.
So, I'd merge it here as a reference implementation and then we can discuss in API repo to give it a broader scope.

matthijskooijman:
> The problem in doing so is that we must ask for a sync.h or similar to be present in the core implementation,

Well, you could just require that a core that uses -API supplies a working `sync.h`, right?

>  So, I'd merge it here as a reference implementation and then we can discuss in API repo to give it a broader scope.

Agreeed.

Issue #570, opened 2020-11-22
Bootloader does not honor security bit


ksmith3036:
When security bit of the SAMD21 MCU is enabled using the command "bossac.exe -s", the bootloader still allows reading the flash memory. 
I think the correct behavour should be for the bootloader to deny read access of the flash when security is enabled, to protect any secrets uploaded to flash as part of a sketch or to protect intellectual property. 

When working with MKR GSM 1400 and MKR NB 1500, the PIN for the SIM card is stored in the flash, and is easily recovered by reading the flash content and do a search for a string of four numbers.

The problem was verified on a MKR GSM 1400, although with an old bootloader from 2017:

C:> arduino15\packages\arduino\tools\bossac\1.7.0-arduino3\bossac.exe -i
Device found on COM5
Atmel SMART device 0x10010005 found
Device       : ATSAMD21G18A
Chip ID      : 10010005
Version      : v2.0 [Arduino:XYZ] Nov 30 2017 10:13:21
Address      : 8192
Pages        : 3968
Page Size    : 64 bytes
Total Size   : 248KB
Planes       : 1
Lock Regions : 16
Locked       : none
**Security     : true**
Boot Flash   : true
BOD          : false
BOR          : false
Arduino      : FAST_CHIP_ERASE
Arduino      : FAST_MULTI_PAGE_WRITE
Arduino      : CAN_CHECKSUM_MEMORY_BUFFER

According to docs from Microchip, the Security Bit isn't active until restart, so the board was reset a number of times, including powering down.

C:> arduino15\packages\arduino\tools\bossac\1.7.0-arduino3\bossac.exe -r c:\temp\flashcontent.bin
Device found on COM5
Atmel SMART device 0x10010005 found

Read 253952 bytes from flash
[==============================] 100% (3968/3968 pages)
done in 1.429 seconds

PR #586, opened 2020-12-22
Make bootloader honour the MCU Security Bit

ksmith3036:
If Security Bit is set, the MCU denies access to most information when conneting using an SWD-connected debugger. One has to perform a Chip Erase to access the MCU.

The default Arduino MKR bootloaders on the other hand, does not honour the Security Bit, and happily lets the bossac.exe program read its flash memory. This might have security or IP implications.

This pull request is a proposed fix for issue #570.

The proposed changes  to the source code will not let a client application read from flash memory, and will always erase the full sketch flash memory, when flashing, to avoid anyone trying to perform a partial flash.

If BOOTPROT is set to 2 and Security Bit is set, the bootloader will be fully protected, and it should also not be trivial to read the sketch out of the MCU. 
Sketches may of course still be flashed into the board.

Protection may also be turned on by setting the SECURE_BY_DEFAULT compile flag. This will not protect against flashing the bootloader itself, but will read-protect the flash storage. 

To allow the MKR VIDOR 4000 loader to build to less than 8 KByte, the string handling of compile time DATE and TIME had to be concatinated in source code.

### DIFF
--- a/bootloaders/zero/Makefile
+++ b/bootloaders/zero/Makefile
@@ -71,10 +71,13 @@ else
   CFLAGS+=-Os -DDEBUG=0 -flto
 endif
 
+ifdef SECURE_BY_DEFAULT
+  CFLAGS+=-DSECURE_BY_DEFAULT=1
+endif
+
 ELF=$(NAME).elf
 BIN=$(NAME).bin
 HEX=$(NAME).hex
-
 
 INCLUDES=-I"$(MODULE_PATH)/tools/CMSIS/4.5.0/CMSIS/Include/" -I"$(MODULE_PATH)/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/"
 
--- a/bootloaders/zero/bootloader_samd21x18.ld
+++ b/bootloaders/zero/bootloader_samd21x18.ld
@@ -27,7 +27,7 @@
 MEMORY
 {
   FLASH (rx) : ORIGIN = 0x00000000, LENGTH = 0x2000 /* First 8KB used by bootloader */
-  RAM (rwx) : ORIGIN = 0x20000000, LENGTH = 0x00008000-0x0004 /* 4 bytes used by bootloader to keep data between resets */
+  RAM (rwx) : ORIGIN = 0x20000000, LENGTH = 0x00008000-0x0400 /* last 4 bytes used by bootloader to keep data between resets, but reserves 1024 bytes for sketches to have same possibility */
 }
 
 /* Linker script to place sections and symbol values. Should be used together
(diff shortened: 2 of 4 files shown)


facchinm:
Hi @ksmith3036 ,
thank you very much for posting this!
I think the rationale is perfect but I have some "style" suggestions.
1. Try keeping the original spaces/tabs formatting so the diff improves significanly (and also the ability to properly review)
2. Use curly braces on single line `if`s (even if C allows this the result is very prone to misread/errors)
3. Split the patch into 3 or more atomic diffs (one could be date/time, the second `b_security_enabled` related bits, the third adding `SECURE_BY_DEFAULT`
Apart from that, well done! I'll be superhappy to merge this once it's fixed!

ksmith3036:
> 1. Try keeping the original spaces/tabs formatting so the diff improves significanly (and also the ability to properly review)

OK, will do. 

> 2. Use curly braces on single line `if`s (even if C allows this the result is very prone to misread/errors)

OK, will do, even if it hurts to split things that syntactically belongs together. :-)  

> 3. Split the patch into 3 or more atomic diffs (one could be date/time, the second `b_security_enabled` related bits, the third adding `SECURE_BY_DEFAULT`

Here I am a bit unsure, because without concatening the DATE and TIME,s the Vidor Bootloader become just over 8 KByte in size.
But, I will comply. 
But I certainly have a problem, because I have no clue about how to do it in Git. Do I revert my changes in some way and then commit each separately. I noticed that on another Git project, when I created a pull request, all my later commits automatically was attached to the same pull request. 
Do the three commits need to be in three different pull requests?

facchinm:
Thank you so much!
So, in git you can do this (from your situation):
```
git reset HEAD~1 # this will "delete" your last commit but all the modification will still be there
git add -i # will launch an interactive shell where you can add pieces of code
# when done, just quit and commit as usual, for every "subcommit"
git push $yourremote master --force # --force rewrites history and keeps this PR up to date (without opening another one)
```
If you feel unsure I can recreate the patch myself, push in your repo for double checking and then we can merge :slightly_smiling_face:

ksmith3036:
Thank you very much for the tips! 
I will try myself, I really need to learn to use Git! :-)

ksmith3036:
Thank you for your advice. 
I split into three commits, hopefully as you want it.  
I added one more commit, as I had forgotten it earlier on. My sketch needs to store a few variables able to survive a reset of the MCU, so I have modifed the Arduino linker script to include a section of 256 bytes of RAM at the end of RAM, but this only works if the bootloader also keeps the stack out of the same area. Because of this, I modifed the bootloader linker script to exclude the last 1 KByte of RAM from use. The bootloader seems happy with this, and really should, having 31 KByte of RAM available! :-)
I hope that also could make it into the official bootloader.

facchinm:
That's perfect! Lovely PR!
I'm merging it now. About distribution, we usally follow this strategy:
* on the next production batch, we'll rebuild the bootloaders and flash them on the boards.
* once the boards have been manufactured / flashed we'll merge the compiled binaries so running "Burn Bootloader" from the IDE will update to something properly validated.

Thank you again!

PR #590, opened 2020-12-25
Slave Wire: fix TwoWire::onService() to handle repeated start

reufer:
Combined I2C write/read transactions with repeated start
conditions were not possible, since only onRequestCallback was
called. Any previously received data was passed to
onReceiveCallback as part of a subsequent master write.

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -219,61 +219,60 @@ void TwoWire::onRequest(void(*function)(void))
 
 void TwoWire::onService(void)
 {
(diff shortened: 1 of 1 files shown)


stonehippo:
@reufer

I don't think this is quite right yet/ The existing onService handler in Wire.cpp seems to have a couple of issues (onRequest and onReceive are backwards, and you have to do two writes to get them to the onReceive callback), and your changes don't fully address them. I've taken a pass at rewriting the whole function and I think I've got it working correctly, including handling repeated starts.

Since you've already take a shot at this, would you mind trying out the changes in #605?

Issue #489, opened 2020-01-21
Only define Serial ISRs when used


matthijskooijman:
(There is an identical issue for sam [here](https://github.com/arduino/ArduinoCore-sam/issues/100))

Currently, I think that the Serial interrupt handlers are always included in the link (or maybe always included if any serial object is used). Ideally, none of these would be included by default, and only the ISRs for serial objects that are actually used are included. This would leave the way open for libraries or sketches to define their own serial ISR for any otherwise unused serial ports.

This behaviour is already implemented on AVR, by putting the definition of each serial object and its corresponding ISR in a separate .cpp file, and separating all serial objects from each other. Due to the way linking works wrt .a files, this should have the desired effect (when care is taken not to include any stray references to these serial objects). See https://github.com/arduino/Arduino/pull/1711#issuecomment-29745866 for the analysis I did on the original implementation on AVR.

I created this issue for sam after suggesting this as an alternative for https://github.com/arduino/ArduinoCore-sam/pull/95 (and see also https://github.com/arduino/ArduinoCore-avr/pull/299), and it made sense to do the same for samd as well.

I will not have time to actually implement this, but I'm happy to support anyone that wants to give a stab at it.

PR #600, opened 2021-02-15
Add a weak attribute to serial IRQ allowing real time communication lib.

nicolas-rabault:
There is no way to have a complexe software communication protocol without IRQ.
Arduino should comply with it allowing real time constraints protocols.

The best way could be to be able to avoid Serial object creation and call a custom object into IRQ, but I can't find any way to do it without breaking everything.

Another good way could be to make something like an attach_interrupt function on Serial object. But it's not really standard into Arduino ecosystem and impact everyone with a little bit more CPU time on each byte reception for the benefits of some specific guys like me.

The last way is to declare IRQ as weak allowing someone to take the lead on IRQ. In this case, there is a dead Serial object but this modification doesn't do anything in any other case and the custom serial lib developper is the guilty one.


matthijskooijman:
This seems like a reasonable approach, but I think we can do better. For AVR, we already set things up so that if `Serial` is not  references, the entire object *and* ISR is omitted. See #489 for brief discussion and links about this approach.

nicolas-rabault:
Thank you for pointing me into this direction.
Indeed it seems to be a better solution, but a more time consuming one to make.
Like you I would love to make it happen but I don't have a lot of time.
Anyway I will make a post-it on my blackboard about it...

Adminius:
@nicolas-rabault hi, any news? I would also like to use my own serial IRQs

Issue #604, opened 2021-02-22
SafeRingBuffer.h compilation warning with 'All'


PhiJu5:
Is it possible to call return in a synchronized block ?

template <int N>
int SafeRingBufferN<N>::read_char() {
  synchronized {
    return RingBufferN<N>::read_char();
  }
}

-> 
template <int N>
int SafeRingBufferN<N>::read_char() {
  int ret = 0;
  synchronized {
    ret = RingBufferN<N>::read_char();
  }
   return ret;
}

At least this fix the compilation warning

Arduino15\packages\arduino\hardware\samd\1.8.11\cores\arduino\USB\SAMD21_USBDevice.h
uint32_t len = 0;
synchronized {
   len = _rx_buffer.availableForStore();
}

facchinm:
Fixed by #618

PR #610, opened 2021-03-09
fix: address warning

zfields:
provide return value for "unused" path


zfields:
Resolves issues #604.

zfields:
I'm scratching my head here...

What is the expected behavior that is being failed by a test? The warning needs to be addressed, because it appears this function can return whatever happens to be on the end of the stack.

Perhaps this needs to be completely refactored by the original author to have a more clear and deterministic outcome.

facchinm:
Replaced by #618, which is a bit cleaner and more commented. Thanks anyway, sorry for the late response

Issue #593, opened 2021-01-14
MKRWAN 1310 Flash Memory access fails.


terryzar:
Using the SerialFlash library, accessing the on board serial flash memory works fine under Arduino SAMD Core library v1.8.10.  After upgrade to 1.8.11 access to the on board flash memory stopped working.   I have not yet had time to analyze this, just rolled back to 1.8.10.

jmbarra78wise:
The same issue for me in MKRGSM1400 plus MKR Mem Shield when upgrade to 1.8.11 so rolled back to 1.8.10.

facchinm:
@terryzar @jmbarra78wise I pushed a fix here https://github.com/arduino/ArduinoCore-samd/pull/622 , would you mind testing it? Thanks!

jmbarra78wise:
It works for me! Thanks for the fix!

PR #622, opened 2021-04-07
SPI: restore 12MHz as maximum frequency

facchinm:
Fixes https://github.com/arduino/ArduinoCore-samd/issues/593

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -94,8 +94,8 @@ void SPIClass::config(SPISettings settings)
     _p_sercom->disableSPI();
 
     uint32_t clock_freq = settings.getClockFreq();
-    if (clock_freq > F_CPU/2) {
-      clock_freq = F_CPU/2;
+    if (clock_freq > SERCOM_FREQ_REF/SPI_MIN_CLOCK_DIVIDER) {
+      clock_freq = SERCOM_FREQ_REF/SPI_MIN_CLOCK_DIVIDER;
     }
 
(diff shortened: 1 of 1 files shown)


matthijskooijman:
Is 12Mhz really the limit of the SPI hardware? Or is this just a limit of the slave devices used in the bug report and/or the wiring used? Also, this says "restore", so it would be good if the commit message would refer to the commit that lifted this limitation previously and comments on why that was wrong (if it is)?

facchinm:
The "previous behaviour" I was referring to is the one documented here https://github.com/arduino/ArduinoCore-samd/blob/master/libraries/SPI/SPI.h#L37-L43 .
It's very likely that the flash under test doesn't support 24Mhz speed, but the library doesn't allow changing the frequency https://github.com/PaulStoffregen/SerialFlash/blob/master/SerialFlashChip.cpp#L33

cmaglie:
I remember writing the comment in SPI.h. Also I remember that someone experimentally determined that the maximum clock is 12Mhz in SPI Master mode.

Looking to an updated datasheet this seems confirmed by the following:
image
image

84ns translates to roughly 1.000.000.000/84=11.904.761 Hz

sambartle:
Do I need to turn off some kind of signature verification to test this? 
The "http://downloads.arduino.cc/PR/samd/package_samd-b212_index.json.sig" file doesn't seem to exist.

sambartle:
Any progress on fixing the signature verification for this so it can be tested? 
It's a really annoying issue causing many problems with other libraries due to their reliance on working SPI.

PR #623, opened 2021-04-07
Small update for WInterrupts.c

Reneg973:
- removed superfluous static variable
- typo fix
- break loop in handler if already found
- some TBD changes

### DIFF
--- a/cores/arduino/WInterrupts.c
+++ b/cores/arduino/WInterrupts.c
@@ -21,16 +21,19 @@
 
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


Reneg973:
Regarding the IRQ register clear I am currently a bit confused. Original code writes a '1' at the IRQ bit to the same register it has been read "to clear the IRQ".
I know there are 2 different ways of resetting an output, first by writing a '0' or second by writing a '1' into an explicitly defined CLEAR register.
But writing a '1' where already '1' was read out is very new to me. Is this an issue or a special behavior? If second, where can I read how this works?

Reneg973:
I'll make a new request when everything is clean

Issue #631, opened 2021-05-21
Make SERCOM(0,1)_Handler optional


maxlem:
Hello,

In file [variant.cpp](https://github.com/arduino/ArduinoCore-samd/blob/master/variants/arduino_zero/variant.cpp#L35), two handlers are defined

```
void SERCOM0_Handler(){  Serial1.IrqHandler();}
void SERCOM5_Handler(){  Serial.IrqHandler();}
```
I understand it simplify the life of many, e.g. but in my case, I use SERCOM5 as an SPI device and I have to comment out that ISR definition.

What I propose is to make this optional by adding define condition

```
#if !defined(_VARIANT_ARDUINO_ZERO_DISABLE_SERCOM_HANDLERS_)
void SERCOM0_Handler(){  Serial1.IrqHandler();}
void SERCOM5_Handler(){  Serial.IrqHandler();}
#endif
```

if it make sense, I will send a PR

maxlem:
> if only the handler is disabled but not the object, then you'd end up with a non-functioning serial object
Yeah that's true, I though so after my comment

So here's another take
```
#if !defined(_VARIANT_ARDUINO_ZERO_DISABLE_SERCOM_HANDLERS_)
Uart Serial1( &sercom0, PIN_SERIAL1_RX, PIN_SERIAL1_TX, PAD_SERIAL1_RX, PAD_SERIAL1_TX ) ;

Uart Serial( &sercom0, PIN_SERIAL1_RX, PIN_SERIAL1_TX, PAD_SERIAL1_RX, PAD_SERIAL1_TX ) ;

void SERCOM0_Handler()
{
  Serial1.IrqHandler();
}

void SERCOM5_Handler()
{
  Serial.IrqHandler();
}

#endif
```

with that version, if you try to use `Serial` you're greeted with
```
main.cpp:(.text.setup+0x28): undefined reference to `Serial'
```
but if you don't, it builds and links just fine (no warnings)

I went to read the other feature request, which I'm all in favor, but it doesn't fix the fact that a SERCOM#_Hanlder() has been defined.

I think both patches are complementary.

Also, can you already send me a CLA, I don't want to wait 16 months!

matthijskooijman:
> I went to read the other feature request, which I'm all in favor, but it doesn't fix the fact that a SERCOM#_Hanlder() has been defined.

[...]

PR #632, opened 2021-05-21
FEAT make Serial and Serial1 def and ISR optional

maxlem:
see also this discussion:
https://github.com/arduino/ArduinoCore-samd/issues/631

### DIFF
--- a/variants/arduino_mzero/variant.cpp
+++ b/variants/arduino_mzero/variant.cpp
@@ -210,14 +210,18 @@ SERCOM sercom3( SERCOM3 ) ;
 SERCOM sercom4( SERCOM4 ) ;
 SERCOM sercom5( SERCOM5 ) ;
 
+#if !defined(ARDUINO_ZERO_CUSTOM_SERCOM0_SERIAL)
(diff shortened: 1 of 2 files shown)


maxlem:
I'm willing to port the defines to the other samd variant.cpp

matthijskooijman:
What's your idea about setting these defines? You can't really set defines from a sketch, so are you thinking about a custom boards.txt that reuses the variant but sets these defines in the extra_flags or something like that?

Looking at the code, I think this now prevents definition of e.g. Serial or Serial1, but not the declaration, leading to linker error. I would think it would make sense to also not declare the objects at all, so you would get a compiler error earlier if you tried to use them, but I'm not quite sure where they are declared exactly.

matthijskooijman:
(Also, I'm just offering some thoughts on the PR since I participated in the issue, I'm not at all sure if adding these guards is the best approach, but I'm also not in a position to decide on whether to merge this or not).

maxlem:
Yes we might just as well not neclare them, I was just shooting for the most minimal changes.

The use case would me to add -D ARDUINO_ZERO_CUSTOM_SERCOM0_SERIAL to the build command.

I don't know how to do that in Arduino IDE, I use platformio

PR #641, opened 2021-07-28
new method enableMasterClock(fsMultiplier)

KillzoneKid:
Outputs MCK on I2S_MCK[0] pin to drive slave peripherals such as PCM1808 ADC for example. Default fs multiplier is 256 which produces 12MHz clock with 48KHz sampling rate, and divides clocks for other channels acordingly. There is also optional fs multiplier argument. The method should be called before 'begin', for example

`
I2S.enableMasterClock();
if (!AudioInI2S.begin(48000, 32))
{
...
`

### DIFF
--- a/libraries/I2S/src/I2S.cpp
+++ b/libraries/I2S/src/I2S.cpp
@@ -28,12 +28,14 @@ static I2SDevice_SAMD21G18x i2sd(*I2S);
 
 int I2SClass::_beginCount = 0;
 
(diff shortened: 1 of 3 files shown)


KillzoneKid:
Seems like AceTime didnt compile,, which is not related to changes in any way. Found this post too https://www.gitmemory.com/issue/arduino/ArduinoCore-samd/606/808888337 seems like it doesnt work well for samd core.

PR #642, opened 2021-07-31
add architecture specific flag to compiler.S.flags

xiongyu0523:
with this PR, .S file does not need to have architecture directives like `.cpu cortex-m0` in the beginning to avoid compiler error for MSR, MRS instructions.


xiongyu0523:
Error reported by CI are not related to this PR. 

See https://forum.arduino.cc/t/please-remove-acetime-from-samd-core-checks/889918/2

facchinm:
LGTM!

PR #651, opened 2021-10-18
UART: restore default buffer size to 256 bytes

### LINKED ISSUE #650: Regression in UART buffer size
In the v1.8.9 release of the core the default ring buffer size was set to 256: https://github.com/arduino/ArduinoCore-samd/blob/1.8.9/cores/arduino/RingBuffer.h#L26-L32

After https://github.com/arduino/ArduinoCore-samd/pull/560 was merged it looks like this is reduced to 64 bytes now as `api/RingBuffer.h` is used instead of `RingBuffer.h` in [cores/arduino/SafeRingBuffer.h](https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/SafeRingBuffer.h): https://github.com/arduino/ArduinoCore-API/blob/master/api/RingBuffer.h#L33

This might also be related to https://github.com/arduino/ArduinoCore-samd/issues/644

facchinm:
Fixes #650

### DIFF
--- a/cores/arduino/Uart.h
+++ b/cores/arduino/Uart.h
@@ -22,7 +22,11 @@
 #include "SERCOM.h"
 #include "SafeRingBuffer.h"
 
-#define SERIAL_BUFFER_SIZE  64
+#ifdef SERIAL_BUFFER_SIZE
+#undef SERIAL_BUFFER_SIZE
+#endif
+
+#define SERIAL_BUFFER_SIZE  256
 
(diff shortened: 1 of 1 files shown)


maxgerhardt:
But then the buffer size is again non-user-overridable?

https://github.com/arduino/ArduinoCore-samd/blob/a5d52dbf7bcf338de2042eb74cfba33993729c81/cores/arduino/Uart.h#L25-L29

PR #665, opened 2022-03-16
added support for new pmic

Rocketct:
added support for MP2629:

- defined a class for manage different chargers
- auto-detect which PMIC is mounted and configure it consequently
- support only for wifi1010

this library is based on https://github.com/arduino/ArduinoCore-samd/pull/569


kpfleming:
Thanks for picking up my work and extending it! There wasn't a CLA requirement when I first opened my PR, but there is now, so I've signed it.

PR #672, opened 2022-06-08
Fixed Serial_::begin(..) definition to match arduino::HardwareSerial

cmaglie:
This is usually not a problem becuase we use the arm-none-eabi-gcc compiler thas has type `uint32_t` aliased to `unsigned long`. Anyway this is not mandatory in general, and the compiler may choose to alias `uint32_t` with `unsigned int` as it happens with llvm/clang.

Since we use clangd as language server on the Arduino IDE 2.0 the Serial_ object and all the derived instances (Serial, SerialUSB, ...) are not available in code-completion suggestions.

### DIFF
--- a/cores/arduino/USB/CDC.cpp
+++ b/cores/arduino/USB/CDC.cpp
@@ -196,12 +196,12 @@ void Serial_::enableInterrupt() {
 	usbd.epBank0EnableTransferComplete(CDC_ENDPOINT_OUT);
 }
 
-void Serial_::begin(uint32_t /* baud_count */)
+void Serial_::begin(unsigned long /* baudrate */)
 {
 	// uart config is ignored in USB-CDC
 }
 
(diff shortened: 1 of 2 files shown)


cmaglie:
For the record this is the "error" detected by clang:

```
In file included from /home/cmaglie/.arduino15/packages/arduino/hardware/samd/1.8.13/cores/arduino/Arduino.h:112:
/home/cmaglie/.arduino15/packages/arduino/hardware/samd/1.8.13/cores/arduino/USB/USBAPI.h:180:16: error: variable type 'Serial_' is an abstract class
extern Serial_ SerialUSB;
               ^
/home/cmaglie/.arduino15/packages/arduino/hardware/samd/1.8.13/cores/arduino/api/HardwareSerial.h:90:18: note: unimplemented pure virtual method 'begin' in 'Serial_'
    virtual void begin(unsigned long) = 0;
                 ^
```

facchinm:
LGTM!

PR #681, opened 2022-09-06
Prevent malloc from overwriting the stack

matthijskooijman:
The current implementation of malloc (or really, the underlying `sbrk()` implementation) happily allocates any amount of memory, even when that overwrites the stack or even beyond the end of RAM. This is fixed by implementing a custom `sbrk()` function that does proper checking. The code is based on the `sbrk()` from the STM32 Arduino core, but with an additional margin added (to make it fail when it comes close to the stack, instead of just when it would actually overwrite the stack).

This margin approach is copied from avr-libc's malloc implementation. I considered also copying more of avr-libc's configurability (e.g. `__malloc_heap_start` and `__malloc_heap_end`), but that ended up just adding complexity without a very clear usecase, so I left that out.

See the d232b594d4673604d0cf360a39dfa66fdef6c4c1 commit message for much more detail on this problem and the solution.

This PR also has a somewhat unrelated commit removing the `-nostdlib` compilation option, which was in the wrong place and thus effectively unused (and also unneeded).

### DIFF
--- a/cores/arduino/main.cpp
+++ b/cores/arduino/main.cpp
@@ -56,3 +56,10 @@ int main( void )
 
   return 0;
 }
+
+// Ensure our version of sbrk is used by forcing a reference to it here
+// (without this, an undefined reference only occurs later when linking
+// newlib, and the dummy from libgloss/nosys will be used instead).
+// This variable will be optimized away later.
+extern "C" void *_sbrk(int);
+void * force_reference_to_sbrk = (void*)&_sbrk;
--- a/cores/arduino/sbrk.c
+++ b/cores/arduino/sbrk.c
@@ -1,0 +1,19 @@
+#include <samd.h>
+#include <stddef.h>
+#include <errno.h>
+
+void *_sbrk(int incr)
+{
+  extern char end; /* End of global variables, defined by the linker */
+  static char *brkval = &end ;
+  char *prev_brkval = brkval;
+
+  if (brkval + incr > (char *)__get_MSP()) {
+    /* Heap and stack collision */
+    errno = ENOMEM;
+    return (void*) -1;
+  }
+
+  brkval += incr ;
+  return (void*) prev_brkval;
+}
--- a/cores/arduino/sbrk.c
+++ b/cores/arduino/sbrk.c
@@ -2,13 +2,17 @@
 #include <stddef.h>
 #include <errno.h>
 
+/* How much free space to keep between the stack and the heap by malloc.
+ * Can be changed by the application at any time. */
+size_t __malloc_margin = 64;
+
 void *_sbrk(int incr)
 {
   extern char end; /* End of global variables, defined by the linker */
   static char *brkval = &end ;
   char *prev_brkval = brkval;
 
-  if (brkval + incr > (char *)__get_MSP()) {
+  if (brkval + incr + __malloc_margin > (char *)__get_MSP()) {
     /* Heap and stack collision */
     errno = ENOMEM;
     return (void*) -1;
--- a/platform.txt
+++ b/platform.txt
@@ -39,13 +39,13 @@ compiler.optimization_flags.debug=-Og -g3
 
 compiler.path={runtime.tools.arm-none-eabi-gcc-7-2017q4.path}/bin/
 compiler.c.cmd=arm-none-eabi-gcc
-compiler.c.flags=-mcpu={build.mcu} -mthumb -c -g {compiler.optimization_flags} {compiler.warning_flags} -std=gnu11 -ffunction-sections -fdata-sections -nostdlib --param max-inline-insns-single=500 -MMD
+compiler.c.flags=-mcpu={build.mcu} -mthumb -c -g {compiler.optimization_flags} {compiler.warning_flags} -std=gnu11 -ffunction-sections -fdata-sections --param max-inline-insns-single=500 -MMD
 compiler.c.elf.cmd=arm-none-eabi-g++
 compiler.c.elf.flags={compiler.optimization_flags} -Wl,--gc-sections -save-temps
 compiler.S.cmd=arm-none-eabi-gcc
 compiler.S.flags=-mcpu={build.mcu} -mthumb -c -g -x assembler-with-cpp -MMD
 compiler.cpp.cmd=arm-none-eabi-g++
-compiler.cpp.flags=-mcpu={build.mcu} -mthumb -c -g {compiler.optimization_flags} {compiler.warning_flags} -std=gnu++11 -ffunction-sections -fdata-sections -fno-threadsafe-statics -nostdlib --param max-inline-insns-single=500 -fno-rtti -fno-exceptions -MMD
+compiler.cpp.flags=-mcpu={build.mcu} -mthumb -c -g {compiler.optimization_flags} {compiler.warning_flags} -std=gnu++11 -ffunction-sections -fdata-sections -fno-threadsafe-statics --param max-inline-insns-single=500 -fno-rtti -fno-exceptions -MMD
 compiler.ar.cmd=arm-none-eabi-ar
 compiler.ar.flags=rcs
 compiler.objcopy.cmd=arm-none-eabi-objcopy


matthijskooijman:
Below is the sketch I used to test this. It just allocates various types of memory using operator new, including some allocations that would overwrite the stack, or even extend past the end of RAM so must fail. There is a bit of disabled code at the end that checks non-no-throw do not return null on failure, but that's a different issue that I'll try to discuss separately later, but I included the test code just FYI.

```
#include <new>

constexpr size_t s = 8000;
constexpr size_t s_fail = 65000;

struct foo {
  char bar[s];
};

struct foo_fail {
  char bar[s_fail];
};

struct C {
  int v;
  C() : v (1234) { Serial.println("    constructor"); Serial.flush(); }
  ~C() { Serial.println("    destructor"); Serial.flush(); }
};

void setup() {
  Serial.begin(115200);
  while (!Serial);
  Serial.println("Starting");
  Serial.flush();

  /***** nothrow new *****/
  Serial.println("Nothrow new");
  Serial.flush();

  {
    char *arr = new (std::nothrow) char[s];
    Serial.print("  arr=0x");
    Serial.print((uintptr_t)arr, HEX);
    Serial.println(" [Expected non-zero]");

    Serial.flush();
    delete[] arr;
  }
  {
    foo *f = new (std::nothrow) foo();
    Serial.print("  f=0x");
    Serial.print((uintptr_t)f, HEX);
    Serial.println(" [Expected non-zero]");
    Serial.flush();
    delete f;
  }
  {
    C *c = new (std::nothrow) C();
    Serial.print("  c=0x");
    Serial.print((uintptr_t)c, HEX);
    Serial.println(" [Expected non-zero]");
    if (c) {
      Serial.print("    c.v=");
      Serial.print(c->v);
      Serial.println(" [Expected 1234]");
    }
    Serial.flush();
    delete c;
  }

  /***** regular new *****/
  Serial.println("Regular new");
  Serial.flush();

  {
    char *arr = new char[s];
    Serial.print("  arr=0x");
    Serial.print((uintptr_t)arr, HEX);
    Serial.println(" [Expected non-zero]");
    Serial.flush();
    delete[] arr;
  }
  {
    foo *f = new foo;
    Serial.print("  f=0x");
    Serial.print((uintptr_t)f, HEX);
    Serial.println(" [Expected non-zero]");
    Serial.flush();
    delete f;
  }
  {
    C *c = new C();
    Serial.print("  c=0x");
    Serial.print((uintptr_t)c, HEX);
    Serial.println(" [Expected non-zero]");
    if (c) {
      Serial.print("    c.v=");
      Serial.print(c->v);
      Serial.println(" [Expected 1234]");
    }
    Serial.flush();
    delete c;
  }

  /***** placement (non-allocating) new *****/
  Serial.println("Placement new");
  Serial.flush();

  {
    char buf[sizeof(C)];

    C *c = new (buf) C();
    Serial.print("  c=0x");
    Serial.print((uintptr_t)c, HEX);
    Serial.println(" [Expected non-zero]");
    if (c) {
      Serial.print("    c.v=");
      Serial.print(c->v);
      Serial.println(" [Expected 1234]");
    }
    Serial.flush();
    c->~C();
  }

  /**** Stack margin test ****/
  Serial.println("Stack margin/overflow tests");
  Serial.flush();

  // End of global variables, defined by the linker, this is
  // SAMD-Arduino-core-specific
  extern char end;
  const size_t room = (char *)__get_MSP() - &end;
  const size_t margin = 64;

  {
    char *arr = new (std::nothrow) char[room - margin - 100];
    Serial.print("  arr=0x");
    Serial.print((uintptr_t)arr, HEX);
    Serial.println(" [Expected non-zero]");
    Serial.flush();
    delete[] arr;
  }

  // Note that (on cores where malloc/sbrk never fails) these might mess
  // up the heap a bit and make future malloc's hardfault for trying to
  // write *after* these free'd invalid allocations
  {
    char *arr_fail = new (std::nothrow) char[room - margin + 1];
    Serial.print("  arr_fail=0x");
    Serial.print((uintptr_t)arr_fail, HEX);
    Serial.println(" [Expected 0]");
    Serial.flush();
    delete[] arr_fail;
  }
  {
    char *arr_fail = new (std::nothrow) char[room + 1];
    Serial.print("  arr_fail=0x");
    Serial.print((uintptr_t)arr_fail, HEX);
    Serial.println(" [Expected 0]");
    Serial.flush();
    delete[] arr_fail;
  }

  /***** nothrow new failing *****/
  Serial.println("Failing allocations");
  Serial.flush();

  // On cores where malloc/sbrk never fails, this might lock up because
  // the heap is already a bit messed up (see above)
  {
    char *arr_fail = new (std::nothrow) char[s_fail];
    Serial.print("  arr_fail=0x");
    Serial.print((uintptr_t)arr_fail, HEX);
    Serial.println(" [Expected 0]");
    Serial.flush();
    delete[] arr_fail;
  }

  // On cores where malloc/sbrk never fails, this will lock up because
  // the allocated memory is zero'd.
  {
    foo_fail *f_fail = new (std::nothrow) foo_fail();
    Serial.print("  f_fail=0x");
    Serial.print((uintptr_t)f_fail, HEX);
    Serial.println(" [Expected 0]");
    Serial.flush();
    delete f_fail;
  }

  # if 0
  // Disabled for now, since most Arduino cores override regular new
  // with a non-aborting version that returns null, which is not
  // standards-compliant, but also breaks struct allocation (which tries
  // to zero out the null pointer and crashes).
  // Reverting to the default gcc operator new versions (that abort
  // instead of returning null) would be preferred, except that gcc
  // 9 and above have no working nothrow versions, so then there is no
  // way to do notrhow allocation at all, see
  // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106477)
  /***** regular new failing *****/

  Serial.print("Next allocation should abort, nothing else should be printed");
  Serial.flush();

  {
    char *arr_fail = new char[s_fail];
    Serial.print("  arr_fail=0x");
    Serial.print((uintptr_t)arr_fail, HEX);
    Serial.println(" [Expected to not return]");
    Serial.flush();
    delete[] arr_fail;
  }

  {
    foo_fail *f_fail = new foo_fail();
    Serial.print("  f=0x");
    Serial.print((uintptr_t)f_fail, HEX);
    Serial.println(" [Expected to not return]");
    Serial.flush();
    delete f_fail;
  }

  Serial.print("This should never print");
  Serial.flush();
  #endif
}

void loop() {}

void HardFault_Handler() {
  static bool printed = false;
  if (!printed) {
    Serial.println();
    Serial.println("HARDFAULT [only printed once]");
    Serial.println();
  }
  printed = true;
  // We return here so interrupts are re-enabled and the print can
  // actually be sent to the UART. Since hardfault has a fixed -1
  // priority, there seems no other way (until the serial code is fixed
  // to allow flushing with interrupts disabled:
  // https://github.com/arduino/ArduinoCore-samd/issues/472)
  //
  // Returning here does mean the code continues and likely triggers
  // a ton more hardfaults.
}
```


With the PR applied, output is as it should be:

```
Starting
Nothrow new
  arr=0x20000BC0 [Expected non-zero]
  f=0x20000BC0 [Expected non-zero]
    constructor
  c=0x20002AF8 [Expected non-zero]
    c.v=1234 [Expected 1234]
    destructor
Regular new
  arr=0x20000BC0 [Expected non-zero]
  f=0x20000BC0 [Expected non-zero]
    constructor
  c=0x20002AF8 [Expected non-zero]
    c.v=1234 [Expected 1234]
    destructor
Placement new
    constructor
  c=0x20007FCC [Expected non-zero]
    c.v=1234 [Expected 1234]
    destructor
Stack margin/overflow tests
  arr=0x0 [Expected non-zero]
  arr_fail=0x0 [Expected 0]
  arr_fail=0x0 [Expected 0]
Failing allocations
  arr_fail=0x0 [Expected 0]
  f_fail=0x0 [Expected 0]
```

Without the patch (samd 1.8.13 downloaded through board manager), this produces incorrect results for the last testcases:

```
Starting
Nothrow new
  arr=0x20000BB0 [Expected non-zero]
  f=0x20000BB0 [Expected non-zero]
    constructor
  c=0x20002AF0 [Expected non-zero]
    c.v=1234 [Expected 1234]
    destructor
Regular new
  arr=0x20000BB0 [Expected non-zero]
  f=0x20000BB0 [Expected non-zero]
    constructor
  c=0x20002AF0 [Expected non-zero]
    c.v=1234 [Expected 1234]
    destructor
Placement new
    constructor
  c=0x20007FCC [Expected non-zero]
    c.v=1234 [Expected 1234]
    destructor
Stack margin/overflow tests
  arr=0x20002AF8 [Expected non-zero]
  arr_fail=0x20002A90 [Expected 0]
  arr_fail=0x20002A50 [Expected 0]
Failing allocations

HARDFAULT [only printed once]
```

The hard fault is because, I think, the stack margin tests messed up the heap. If you disable those, the hard fault is delayed one testcase (but still occurs because the next allocation is a struct that is automatically zero'd, so the `memset` call generated by gcc ends up writing beyond the end of RAM).

PR #686, opened 2022-10-14
Fix typo in startup.c comment

### LINKED ISSUE #685: Typo in startup.c comment
I think it's supposed to say 32.768kHz here: https://github.com/arduino/ArduinoCore-samd/blob/89833081ef18aea430b8943ab653dbbd510d43ce/cores/arduino/startup.c#L30

I know you can figure this out from the data sheet but since this modifies the chip's defaults it should be a little more clear, it confused me at first and might save other people time.

ayoung19:
Fixes #685 the clock speed typo in the comments, should make it clearer for future readers.


henklaak:
Why are we not progressing on this kind of trivial but quality contributions?

Issue #451, opened 2019-09-16
Incompatibility between MKRZERO RTC & I2S libraries


white-ar:
There appears to be an incompatibility between the ArduinoSound.h and RTCZero.h libraries.  

When you try to play an audio file after the internal RTC has been initialised, the WAV files won't play. See the `rtc.begin()` function call in the code below.
Configuration: MKRZERO and Adafruit MAX98357A I2S audio board.  

```c++
#include <SD.h>               // SD card access library
#include <ArduinoSound.h>     // I2S sound interface library 
#include <RTCZero.h>          // Real Time Clock library


// SETUP Audio playback
// filename of wave file to play
const char filename[] = "MUSIC.WAV";
// variable representing the Wave File
SDWaveFile waveFile;

// SETUP Real Time Clock
/* Create a RTC object */
RTCZero rtc;

/* Change these values to set the current initial time */
const byte seconds = 0;
const byte minutes = 0;
const byte hours = 16;

/* Change these values to set the current initial date */
const byte day = 15;
const byte month = 6;
const byte year = 15;

void setup() {
  // put your setup code here, to run once:
  Serial.begin(9600);     // Start Arduino serial port at 9,600 baud
  //SD.begin();             // Open the SD card
  delay(500);

  while (!Serial) {
    ; // wait for serial port to connect. Needed for native USB port only
  }

  // setup the SD card, depending on your shield of breakout board
  // you may need to pass a pin number in begin for SS
  Serial.print("Initializing SD card..z.");

  if (!SD.begin()) {
    Serial.println("initialization failed!");
    return;
  }
  Serial.println("initialization done.");

  // create a SDWaveFile
  waveFile = SDWaveFile(filename);

  // check if the WaveFile is valid
  if (!waveFile) {
    Serial.println("wave file is invalid!");
    while (1); // do nothing
  }

  // print out some info. about the wave file
  Serial.print("Bits per sample = ");
  Serial.println(waveFile.bitsPerSample());

  long channels = waveFile.channels();
  Serial.print("Channels = ");
  Serial.println(channels);

  long sampleRate = waveFile.sampleRate();
  Serial.print("Sample rate = ");
  Serial.print(sampleRate);
  Serial.println(" Hz");

  long duration = waveFile.duration();
  Serial.print("Duration = ");
  Serial.print(duration);
  Serial.println(" seconds");

  // adjust the playback volume
  AudioOutI2S.volume(100);

  // check if the I2S output can play the wave file
[...]

PR #692, opened 2023-01-21
Update I2S.cpp to work with RTCZero library

Khosos:
I2S.cpp configures the clock without specifying what to GLCK->DIVSEL bit. This causes issues when the bit is set to 1 by other libraries such as RTCZero. Explicitly setting this value fixes the problem. More can be found about this issue in the following thread https://github.com/arduino/ArduinoCore-samd/issues/451


facchinm:
MKRGSM failure in CI is due to [this](https://github.com/arduino-libraries/MKRGSM/blob/master/examples/Tools/FileUtilsJsonFile/FileUtilsJsonFile.ino) unrelated example.

Merging the patch, thanks @Khosos  for the contribution!

PR #717, opened 2024-06-14
Remove eAnalogReference from ArduinoCore-samd

rubenstar:
Currently, the header for `analogReference(...)` is defined in ArduinoCore-API ([api/Common.h](https://github.com/arduino/ArduinoCore-API/blob/master/api/Common.h#L100)).

My proposal is to move this enum to ArduinoCore-API and align the naming to the scheme that is used there. 

This comes with the additional benefit of fixing a bug where the header that is defined over at `ArduinoCore-API` takes `uint8_t` as argument type, which leads to a compilation warning when compiling together with `ArduinoCore-samd`.

I filed a corresponding PR there too:
https://github.com/arduino/ArduinoCore-API/pull/236

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -73,18 +73,6 @@ extern "C"{
 #ifdef __cplusplus
 extern "C" {
 #endif
-/*
- * \brief SAMD products have only one reference for ADC
- */
-typedef enum _eAnalogReference
-{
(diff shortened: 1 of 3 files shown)


rubenstar:
This unfortunately breaks a lot of the sketches as the corresponding change has not yet landed in `ArduinoCore-API` :(

